PR #8, opened 2017-08-21
active flag for stats API

tarunsinghal92:


llaske:
Looks nice but in my opinion it's better to not check for stats active in delete and findAll else it will force to adapt graphs to react to this new error.
I think you could always retrieve existing stats even if it's not active. You could also remove some existing stats if you like. The only thing you can't do is add new stat to avoid add more lines in DB.

tarunsinghal92:
changed!

llaske:
What if an admin call a request for this user via the dashboard ? The update time should be updated only if the user himself is connected

tarunsinghal92:
The time is updated based on the "x-key" header which is the User_id of the user logged in.

tarunsinghal92:
👍

PR #26, opened 2019-02-23
#19 fixed

ashish0910:

PR #32, opened 2019-02-24
create new user with space in name

ashish0910:
This will allow to create users with spaces in their name


llaske:
Seems that it duplicate https://github.com/llaske/sugarizer-server/pull/26

ashish0910:
Yes but that time it was for classroom and this time it is for users . Both had same problem .

PR #34, opened 2019-02-25
[REFACTOR] Restructed Activities Controllers

imdhruvgupta:


imdhruvgupta:
Hi Tarun,
Please review. I have removed the unwanted binary files.
Thanks,

screenshot 2019-02-25 at 9 47 00 pm

screenshot 2019-02-25 at 9 47 12 pm

dashboard/route.js
```diff
@@ -1,7 +1,7 @@
 // require files
 var authController = require('./controller/auth'),
 	usersController = require('./controller/users'),
-	activitiesController = require('./controller/activities'),
+	activitiesController = require('./controller'),
```
tarunsinghal92: this looks wrong. it should be `require('./controller/activities'),`

dashboard/controller/index.js
```diff
@@ -0,0 +1,6 @@
+export {
+    fakeLaunch,
+    init,
```
tarunsinghal92: this can have problems when you will add other controllers.

tarunsinghal92:
@imdhruvgupta Please check new comments.

imdhruvgupta:
> @imdhruvgupta Please check new comments.

@tarunsinghal92 Yes, I did not realised that. Please review.
Thanks

dashboard/controller/activities/util/index.js
```diff
@@ -0,0 +1,48 @@
+// include libraries
+var request = require('request'),
+    common = require('../helper/common');
+
+function getActivity(req, aid, callback) {
```
tarunsinghal92: why does `getActivity` is inside utils?

imdhruvgupta: The functions defined in utils are being used only in launch and fakeLaunch functions and are private to activity controllers. So, to avoid writing redundant code, I put them inside utils and then imported them from there, so that it can be used in launch and fakeLaunch controllers. 
What are your suggestions?

> why does `getActivity` is inside utils?

tarunsinghal92: @imdhruvgupta: okay,  makes sense.
Could you also look at the `.DS_Store` file. I can still that in recent commit

imdhruvgupta: @tarunsinghal92 There were multiple .DS_Store files. I have removed them all now.

NikhilM98:
@imdhruvgupta @tarunsinghal92 there are some issues which are causing breaking changes. ES6 syntax is used here which is not recognized by node until we use Babel. I think there's no need to use Babel in this project for now. This is causing breaking issues in the project. After pulling the latest code. I'm getting the error:

	> node sugarizer.js

	FATAL ERROR, uncaught exception 'Unexpected token export'
	/sugarizer-server/dashboard/controller/activities/index.js:5
	export { fakeLaunch } from "./fakeLaunch";
	^^^^^^

	SyntaxError: Unexpected token export

PR #40, opened 2019-02-28
Updated README

usmanmuhd:
Updated the image in README with:
dashboard_capture1


llaske:
We don't need to change it now. Some others change will come very soon.

PR #43, opened 2019-03-01
Moving moment translations to front-end

### LINKED ISSUE #42: Language switching not handled properly
Some text on the dashboard is not being translated on switching the language using the language dropdown.

Steps to reproduce:
1. Go to Dashboard Login and login with admin credentials.
screenshot from 2019-03-01 23-17-14

2. After login. Switch language using the dropdown on the top right of the page.
screenshot from 2019-03-01 23-17-27

3. Notice that after changing the language. The "# of Entries" and timestamp is not translated.
screenshot from 2019-03-01 23-17-47

4. Now change the url. Set lang=en and press enter.
screenshot from 2019-03-01 23-17-55

Boom! The text is now translated.

I would like to work on this issue. I'll submit a PR as soon as I find a fix :hammer:

NikhilM98:
Timestamp and some other things are not translated front the front-end. I have made a quick fix. I passed the language parameter in the URL and reloaded the page (Thus requesting appropriate translations from the server).

Closes #42


tarunsinghal92:
@NikhilM98 Could you please attach some screenshots from testing.

NikhilM98:
Here are the screenshots showing the language change when I switch language from English to French using the dropdown. As you can see, the url is changed correctly and the remaining text is now translated.
screenshot from 2019-03-02 10-46-38
screenshot from 2019-03-02 10-46-45

ashish0910:
Looks like some bugs still persists

screenshot 2019-03-02 at 5 08 46 pm

screenshot 2019-03-02 at 5 11 07 pm

NikhilM98:
Thank you @ashish0910 for taking the time to test my implementation.

The issue you found is related to missing translations in `locales.ini` file.
screenshot from 2019-03-02 17-23-39

I think translations for these languages can be added in another pull request. I can handle the English-Hindi and English-French translations.

tarunsinghal92:
@NikhilM98 

Could you also add the missing translation in this pr?

I will review both of them together.

Thanks,

NikhilM98:
@tarunsinghal92 but I already created another PR for language change 8 hours ago [here](https://github.com/llaske/sugarizer-server/pull/48).

dashboard/public/js/main.js
```diff
@@ -260,6 +260,7 @@ function onLocalized() {
 		lang.onchange = function() {
 			l10n.setLanguage(this.value);
 			localStorage.setItem("languageSelection", this.value);
+			location.href = window.location.pathname + "?lang="+document.getElementById('languageSelection').value;
```
tarunsinghal92: @NikhilM98 

NO REFRESH!
This defeats the purpose of changing the language without refresh. Moreover, I didn't see this issue on dev.

NikhilM98: I compared the files of dev and master. They have same issue.

Dev:
screenshot from 2019-03-07 17-47-11

Master:
screenshot from 2019-03-07 17-47-04

(Some of the text refrenced in this issue was not translated due to missing translations. The commit refrencing missing translations been already merged.)

The timestamp is not being translated because `moment.locale` defined in the server is used for translating timestamp. That's why I recommended refreshing the page. But now I understand that refreshing the page is a bad choice.

https://github.com/llaske/sugarizer-server/blob/a56dd9ac441df60faa9628778c69c07249eb5006/dashboard/controller/journal.js#L19-L22

I think getting timestamp language from the server is a also not a good idea. I can see a hack here. Since only one instance of moment is initialized in the project. A user can change the timestamp language for other users.

I'd suggest setting `moment.locale()` in front-end insted of back-end. I also recommend updating minified version. Current minified version is "2.5.1". There are major changes in moment.js after ["2.8.0"](https://github.com/moment/moment/blob/develop/CHANGELOG.md#280-see-changelog) which can be relevent to us.

What do you say @tarunsinghal92?

NikhilM98: Update: I'm trying to translate timestamp in front-end. You can see the progress [here](https://github.com/NikhilM98/sugarizer-server/commit/28bb44ab1d38d20a2c39cbb96e1b652b66a23238).
I'd to made more changes than expected. Please check the diff of last commit on `nikhilm98/moment-update` and review.

NikhilM98: I've not handled moment translations for the timestamps in the charts. The data for the charts is received from the backend, so I think we can modify the API for the charts. We can pass the language in the call, translate in the backend and receive translated data.

ahmedkrmn:
@NikhilM98 Nice. I've just faced this issue and tested your PR here.

There are still some issues:
- When you change the language, Admin and Student in the search bar are not translated to their equivalent. 
Screenshot (64)

- It's also the same for classrooms. Try switching the language to Spanish and reload the page, it works, 
Screenshot (65)


- but revert back to English and you'll notice this, only the first entry in the drop down is translated 
Screenshot (66)

NikhilM98:
@ahmedkrmn actually the scope of this PR is towards handling moment translations on the client side.
This PR already has enough file changes. Adding more fixes would make it difficult for the maintainers to review.

You did a great job finding these issues. I admit that I couldn't really spot these missing translations. 
If you like, you can go ahead and make a PR fixing the issues you mentioned.

llaske:
Hmmm. Lot of change for few things. 
A simple fix is just to force the reload the page when user language is change.

NikhilM98:
@llaske :cold_sweat: that's exactly what I did earlier. But since I was told.

> NO REFRESH!

I took "NO REFRESH!" as a challenge to me and tried to handle it that way.

llaske:
You win the challenge 🏆 Congratulations! 😃 
BTW in this specific case, refresh seems acceptable for me.
Plus all language stuff is handle on the server side, it has no sense to put only date formatting on the client part.

NikhilM98:
Thanks @llaske
But there's an issue with translating in the backend. 
> I think getting timestamp language from the server is also not a good idea. I can see a hack here. Since only one instance of moment is initialized in the project. A user can change the timestamp language for other users.

The similar thing holds true for the notifications and charts translated on the backend. That's why I recommended moving the translations to the front end.
In case you prefer translating in the backend, I suggest sending language as a header in each request so that the server can more accurately identify the language to respond in. (We still have to refresh in this case)

PR #45, opened 2019-03-02
Use single instance of db connection

NikhilM98:
`client.connect` was not handled in the `api/controller`. If the database is not running, the application should exit with an error code and return that application cannot be connected with DB.

### DIFF
--- a/api/controller/classrooms.js
+++ b/api/controller/classrooms.js
@@ -8,19 +8,9 @@ var db;
 var classroomsCollection;
 
 // Init database
-exports.init = function(settings, callback) {
+exports.init = function(settings, database) {
   classroomsCollection = settings.collections.classrooms;
-  var client = new mongo.MongoClient(
-	  'mongodb://'+settings.database.server+':'+settings.database.port+'/'+settings.database.name,
-	  {auto_reconnect: false, w:1, useNewUrlParser: true});
-
-  // Open the db
-  client.connect(function(err, client) {
-	  db = client.db(settings.database.name);
-    if (err) {
-    }
-    if (callback) callback();
-  });
+  db = database;
 };
 
 /**
--- a/api/controller/journal.js
+++ b/api/controller/journal.js
@@ -12,46 +12,36 @@ var shared = null;
 //- Utility functions
 
 // Init database
-exports.init = function(settings, callback) {
+exports.init = function(settings, database) {
 
 	// Open the journal collection
 	journalCollection = settings.collections.journal;
-	var client = new mongo.MongoClient(
-		'mongodb://'+settings.database.server+':'+settings.database.port+'/'+settings.database.name,
-		{auto_reconnect: false, w:1, useNewUrlParser: true});
 
-	// Open the db
-	client.connect(function(err, client) {
-		db = client.db(settings.database.name);
-		if (!err) {
-			db.collection(journalCollection, function(err, collection) {
-				// Get the shared journal collection
-				collection.findOne({
-					'shared': true
-				}, function(err, item) {
-					// Not found, create one
-					if (!err && item == null) {
-						collection.insertOne({
-							content: [],
-							shared: true
-						}, {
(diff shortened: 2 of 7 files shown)


NikhilM98:
Before handling else:
screenshot from 2019-03-04 02-21-09

After handling else:
screenshot from 2019-03-04 01-07-32

Now its returning that it cannot connect with the DB and exits the process with -1 error code.

NikhilM98:
Yes, I was also wondering why DB connections were defined redundantly in each `api/controller`. It was establising multiple connections with the DB. I was new to the repository back then.
Now when I think about it. I can think of two things:
- The function `waitConnection` can be modified to call the callback with `client` as an argument (Without calling `client.close()`).The client would then be passed down to `./api/route` and then further passed down to different routes (Using `.init`).
- We can also export the `client` from `wait4db.js` and it can be imported and used by the `api/controller` for DB queries. 
https://github.com/llaske/sugarizer-server/blob/7d54cf1239929513abcd533362a5e68663870c60/config/wait4db.js#L14-L37

@tarunsinghal92 which approach do you think we should take?

tarunsinghal92:
> Yes, I was also wondering why DB connections were defined redundantly in each `api/controller`. It was establising multiple connections with the DB. I was new to the repository back then.
> Now when I think about it. I can think of two things:
> 
> * The function `waitConnection` can be modified to call the callback with `client` as an argument (Without calling `client.close()`).The client would then be passed down to `./api/route` and then further passed down to different routes (Using `.init`).
> * We can also export the `client` from `wait4db.js` and it can be imported and used by the `api/controller` for DB queries.
>   
>     
>       
>         [sugarizer-server/config/wait4db.js](https://github.com/llaske/sugarizer-server/blob/7d54cf1239929513abcd533362a5e68663870c60/config/wait4db.js#L14-L37)
>       
>       
>           Lines 14 to 37
>         in
>         [7d54cf1](/llaske/sugarizer-server/commit/7d54cf1239929513abcd533362a5e68663870c60)
>       
>       
>       
>       
>   
>           
>             
>              exports.waitConnection = function(settings, callback) { 
>           
>   
>           
>             
>              	var waitTime = settings.database.waitdb; 
>           
>   
>           
>             
>              	if (waitTime) { 
>           
>   
>           
>             
>              		var timer = setInterval(function() { 
>           
>   
>           
>             
>              			server = new Server(settings.database.server, settings.database.port, {auto_reconnect: false}); 
>           
>   
>           
>             
>              			db = new Db(settings.database.name, server, { 
>           
>   
>           
>             
>              				w: 1 
>           
>   
>           
>             
>              			}); 
>           
>   
>           
>             
>               
>           
>   
>           
>             
>              			// Open the db 
>           
>   
>           
>             
>              			db.open(function(err, db) { 
>           
>   
>           
>             
>              				if (!err) { 
>           
>   
>           
>             
>              					clearInterval(timer); 
>           
>   
>           
>             
>              					db.close(); 
>           
>   
>           
>             
>              					callback(); 
>           
>   
>           
>             
>              				} else { 
>           
>   
>           
>             
>              					console.log("Waiting for DB..."); 
>           
>   
>           
>             
>              				} 
>           
>   
>           
>             
>              			}); 
>           
>   
>           
>             
>              		}, waitTime*1000); 
>           
>   
>           
>             
>              	} else { 
>           
>   
>           
>             
>              		callback(); 
>           
>   
>           
>             
>              	} 
>           
>   
>           
>             
>              }
> 
> @tarunsinghal92 which approach do you think we should take?

make the db connection and pass it to controller while initializing them

NikhilM98:
Passed the tests.
Screenshot from 2019-03-11 00-59-36

NikhilM98:
@llaske @tarunsinghal92 can you review this PR. The objective of this PR is to reduce the number of active mongodb instances the server is using.

NikhilM98:
@tarunsinghal92 rebased with upstream dev

PR #47, opened 2019-03-02
[REFACTOR] Restructured graph controllers

imdhruvgupta:


NikhilM98:
@imdhruvgupta I think you've installed Babel globally. That's why the use of ES6 syntax is not causing any issues while testing. Destructuring is not recognized by es2015. So it will not run for others. You need to convert the code in es2015 format.

imdhruvgupta:
@tarunsinghal92 Please review.
screenshot 2019-03-08 at 10 29 33 pm

PR #58, opened 2019-03-04
Installing linter in the project

NikhilM98:
@llaske @tarunsinghal92 I have written the configuration for eslint. I've tried to minimize the errors.
Currently, it's detecting 278 errors. Out of which 9 can be automatically fixed. Should I go ahead and fix those errors manually?

Also, I need suggestions for linting `.ejs` files. I could not find any good plugin for linting them.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,17 @@
+dashboard/public/js/bootstrap-notify.js
+dashboard/public/js/bootstrap.min.js
+dashboard/public/js/Chart.min.js
+dashboard/public/js/jquery-3.1.0.min.js
+dashboard/public/js/jquery-sortable.js
+dashboard/public/js/jquery-ui-1-11-4.js
+dashboard/public/js/jquery.multi-select.js
+dashboard/public/js/jquery.searchable.js
+dashboard/public/js/l10n.js
+dashboard/public/js/material-dashboard.js
+dashboard/public/js/material.min.js
+dashboard/public/js/moment.min.js
+dashboard/public/js/noty.js
+dashboard/public/js/pace.min.js
+dashboard/public/js/select2.min.js
+
+docs/www/vendor/
--- a/package.json
+++ b/package.json
@@ -31,13 +31,15 @@
 		"chai-http": "*",
 		"nodemon": "*",
 		"apidoc": "*",
-		"mocha": "*"
+		"mocha": "*",
+		"eslint": "^5.15.0"
 	},
 	"scripts": {
 		"test": "mocha api/test/",
 		"docs": "apidoc -i api/controller -i dashboard/helper -o docs/www/",
 		"start": "node sugarizer.js",
 		"start:watch": "nodemon sugarizer.js",
-		"seed:users-csv": "node scripts/seed_users.js"
+		"seed:users-csv": "node scripts/seed_users.js",
+		"lint": "eslint .; exit 0"
 	}
 }
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,91 @@
+{
+    "env": {
+        "browser": true,
+        "node": true,
+        "mocha": true,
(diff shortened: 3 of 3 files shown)


tarunsinghal92:
@NikhilM98 

Could you categorize the errors? We can then discuss about the resolution.

NikhilM98:
@tarunsinghal92 The majority of errors can be categorized and listed in the order of priority as:
- Undefined Variables.
- Variables already defined.
- Variable defined but never used.
- Unnecessary/Missing Semicolon.
- Mixed spaces and tabs.
- Indentation. Linebreaks.

llaske:
Look like minor errors.

NikhilM98:
@llaske the errors are minor but if we merge the linter configuration then it will detect 8250 such minor errors and project will appear red with errors.
If I fix all those errors at once. It will be difficult to review the changes and the git history won't look clean.
- Fix all errors at once. Difficult to review.
- Handle one type of errors in the project in one commit. Easier to review.
- Handle batches of files at once. Like fixing api-controller in one go, dashboard-controller in the next. It will also be easier to review and keep the git history clean.

tarunsinghal92:
`Handle one type of errors in the project in one commit` Don't push more till the previous one got approved.

llaske:
I'm not in favor to implement this feature now.
It require too much change in current source code without a real benefit.

NikhilM98:
@llaske too many changes are the reason why I wanted to discuss the implementation with you.

I think we can install the linter with only basic configuration to lint errors related to the undefined, defined but not used and use of es6/es7 syntax. It can be achieved with very less changes in the code.

As the project will grow larger, someday or the other we've to integrate linter, codacy, travis, etc. in this project. Installing linter with basic configuration can be the first step towards that. Not having coding conventions is also a barrier towards attracting contributions.

.eslintignore
```diff
@@ -0,0 +1,17 @@
+dashboard/public/js/bootstrap-notify.js
+dashboard/public/js/bootstrap.min.js
+dashboard/public/js/Chart.min.js
+dashboard/public/js/jquery-3.1.0.min.js
+dashboard/public/js/jquery-sortable.js
+dashboard/public/js/jquery-ui-1-11-4.js
+dashboard/public/js/jquery.multi-select.js
+dashboard/public/js/jquery.searchable.js
```
tarunsinghal92: `cant you just use dashboard/public/*`

package.json
```diff
@@ -31,13 +31,15 @@
...
+		"eslint": "^5.15.0"
 	},
 	"scripts": {
 		"test": "mocha api/test/",
 		"docs": "apidoc -i api/controller -i dashboard/helper -o docs/www/",
 		"start": "node sugarizer.js",
 		"start:watch": "nodemon sugarizer.js",
-		"seed:users-csv": "node scripts/seed_users.js"
+		"seed:users-csv": "node scripts/seed_users.js",
+		"lint": "eslint .; exit 0"
```
tarunsinghal92: why do we need `exit 0`

NikhilM98: If ESlint detects errors in the file, it exits with error code 1. The output looks like the node script is crashing. I've used `exit 0` to suppress the error.

.eslintignore
NikhilM98: We need to lint files such as `main.js` which is stored in the same directory.

P.S. I guess we can further restructure the project to separate public resources with project js files

package.json
tarunsinghal92: yeah you can format this file https://github.com/llaske/sugarizer-server/blob/master/dashboard/public/js/main.js and move code to respective controllers.

.eslintignore
tarunsinghal92: yeah you can format this file https://github.com/llaske/sugarizer-server/blob/master/dashboard/public/js/main.js and move code to respective controllers.

PR #53, opened 2019-03-03
sort added in user and classroom

ashish0910:
issue #50 fixed
Sort by name

Sort by created time

PR #67, opened 2019-03-06
Add Size column in journal view

ashish0910:
sorting added in #53 

Test screenshot

screenshot 2019-03-06 at 9 41 16 am


dashboard/views/journal.ejs
```diff
@@ -98,6 +99,7 @@
 																	</script>
 																	<td><%= entries.entries[i].metadata.title %></td>
 																	<td><%= moment(entries.entries[i].metadata.timestamp).calendar() %></td>
+																	<td><%= entries.entries[i].metadata.file_size %></td>
```
llaske: The right field is not `file_size` but `textsize`.

llaske:
You've use the wrong field (see my review).
Plus, could you format the value using a function similar to the one in Sugarizer [here](https://github.com/llaske/sugarizer/blob/dev/lib/util.js#L543).
Please, do a new PR for it.

ashish0910:
@llaske done in #94

PR #70, opened 2019-03-07
Convert classrooms and users to csv

ashish0910:
Creates csv file of users and classrooms views and download it on client side .

@tarunsinghal92 please let know what other data values need to included in the csv file for both users and classroom.

current data columns for user - name , language  , role
current data columns for classroom - name , student count

ezgif com-video-to-gif

### DIFF
--- a/dashboard/views/classrooms.ejs
+++ b/dashboard/views/classrooms.ejs
@@ -18,7 +18,8 @@
                         </div>
                         <div class="col-md-3"></div>
                         <div class="col-md-6">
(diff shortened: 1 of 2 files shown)


llaske:
Not sure we need that.
We're planning to do new features to import/export classroom/users from CSV files.

ashish0910:
@llaske yes I am planning to add this feature , but first I have added the export feature but wanted some remarks what changes can be done so that I could work further on it .

ashish0910:
> It'll not fetch all the classrooms due to pagination
Yes you are right . We need to incorporate this feature using backend , frontend method like this will not work here because of pagination

PR #72, opened 2019-03-07
Added missing translations

NikhilM98:
Some translations were previously missing. I've added those translations.
Also fixed an error related to passing arguments in file `dashboard/views/classrooms.ejs`


llaske:
Nice. Thanks.

PR #57, opened 2019-03-04
Classroom and User page crash #56 fixed

ashish0910:
Fixed #56 

screenshot 2019-03-04 at 3 38 33 pm

Student count increased 

screenshot 2019-03-04 at 3 46 55 pm

PR #75, opened 2019-03-08
[FIX] Creating classroom with single student

NikhilM98:
When I add a classroom with a single student, a string is stored in the DB instead of an array.

Steps to reproduce the  error:
1. Login to the dashboard. Click on `Classrooms` tab.
2. Click on `ADD CLASSROOM` button and create a classroom with one student.
3. If you check the database, you'll find a string insted of an array in the classrooms collection. If you list the classrooms, count for that classroom will be incorrect and `view students` button will be disabled.

Current Behavior:
screenshot from 2019-03-08 18-55-01
(See the last classroom. Student count is incorrect and we view students button is disabled.)

Expected Behavior:
screenshot from 2019-03-08 18-58-38

P.S. Also removed a function defined inside a function. The scope of the function was incorrect and it was unnecessary.

### DIFF
--- a/dashboard/controller/classrooms.js
+++ b/dashboard/controller/classrooms.js
@@ -38,7 +38,7 @@ exports.index = function(req, res) {
 		query['offset'] = req.query.offset;
 	}
 
-    // call
+	// call
 	request({
 		headers: common.getHeaders(req),
 		json: true,
@@ -73,6 +73,9 @@ exports.addClassroom = function(req, res) {
 		// validate
 		req.body.name = req.body.name.trim();
 		req.body.students = req.body.students || [];
+		if (typeof req.body.students == 'string') {
+			req.body.students = [req.body.students];
+		}
 		req.body.color = JSON.parse(req.body.color);
 		req.assert('name', common.l10n.get('UsernameInvalid')).matches(/^[a-z0-9 ]+$/i);
 		req.body.options = { sync: true, stats: true };
@@ -136,8 +139,8 @@ exports.editClassroom = function(req, res) {
 			req.body.name = req.body.name.trim();
 			req.body.students = req.body.students || [];
 			
-			if (isString(req.body.students)) {
-				req.body.students = [req.body.students]
+			if (typeof req.body.students == 'string') {
+				req.body.students = [req.body.students];
 			}
 
 			req.body.color = JSON.parse(req.body.color);
@@ -205,9 +208,6 @@ exports.editClassroom = function(req, res) {
 			msg: common.l10n.get('ThereIsError')
 		});
 		return res.redirect('/dashboard/classrooms');
-	}
-	function isString(arg) {
-		return typeof arg === 'string';
 	}
 };
 


ashish0910:
@NikhilM98 looks like already solved in #57

NikhilM98:
@ashish0910 actually it was not fixed completely. The issue is still present in `dev`. You can check the files changed tab for details.

NikhilM98:
@ashish0910 actually it was not fixed completely. It's still present in upstream `dev`. You can check the `Files changed` tab to view the the diff.

ashish0910:
From file changes it seems it already fixed 

https://github.com/llaske/sugarizer-server/blob/be03d5e7a79f5571e56e1bd31acaf7882a1eacef/dashboard/controller/classrooms.js#L139

if statement uses this function to check for string 
https://github.com/llaske/sugarizer-server/blob/be03d5e7a79f5571e56e1bd31acaf7882a1eacef/dashboard/controller/classrooms.js#L209

llaske:
It's fixed for me too. Could you check that you've got the latest version of dev.

NikhilM98:
Not completely fixed. It's still present here
https://github.com/llaske/sugarizer-server/blob/be03d5e7a79f5571e56e1bd31acaf7882a1eacef/dashboard/controller/classrooms.js#L74-L76

dashboard/controller/classrooms.js
```diff
@@ -136,8 +139,8 @@ exports.editClassroom = function(req, res) {
 			req.body.name = req.body.name.trim();
 			req.body.students = req.body.students || [];
 			
-			if (isString(req.body.students)) {
-				req.body.students = [req.body.students]
+			if (typeof req.body.students == 'string') {
```
tarunsinghal92: Could you please look into the reason why`req.body.students ` is coming out to be a string.

NikhilM98: It was quite difficult to pinpoint the issue.

I first tested jquery-multiselect, it was storing single element correctly - in an array.

Then I tested the from. The data is submitted by using traditional form data instead of json data submitted by using fetch/request api.

When I used jquery multiselect to select two elements and submit data using form method='POST':
Screenshot from 2019-03-10 21-49-10

When I selected single element and submitted the form method='POST':
Screenshot from 2019-03-10 21-49-34

As you can see the way form data is submitted, it's not possible for the server parser to identify that element with unique key in the request is an array or not.

llaske:
@NikhilM98 could you update unit test to add a test with one student only. Currently test is only with two students. It will allow to cover this case.

NikhilM98:
@llaske actually I'm not really sure that adding test would be much helpful.

Reason:
- When we run tests, we're testing for the 'routes' in 'api/controller'. If you take a look at the `addClassroom` function, you'll notice that it doesn't have validation and it inserts anything that comes from the request into the DB:
https://github.com/llaske/sugarizer-server/blob/766eda390bf2c58c7d8948a21cc01912fb9130c1/api/controller/classrooms.js#L61-L97
It won't be possible to detect this issue that way.

- The validation part is written in 'dashboard/controller'. That's where the issue is present. 
https://github.com/llaske/sugarizer-server/blob/766eda390bf2c58c7d8948a21cc01912fb9130c1/dashboard/controller/classrooms.js#L69-L82
The response of `addClassroom` in this case is an HTML document so even if we write a test for this 'dashboard/controller' export, testing it would be difficult. 

- The test is written with two students so that to validate that only the correct student is removed. If there were only one student then it wouldn't be possible to check that only the correct student is removed.

I can add a test for creating a classroom with a single student but it would be pointless since the test for creating the classroom with two students is testing exactly the same unit.

llaske:
Ok, you're right, it has no sense here.

PR #77, opened 2019-03-10
sorting added 

ashish0910:
This for adding sorting in user , classroom and journal views . 
This is same as #53 with added remarked changes . previous pr closed due to merge conflicts . 
@llaske done the changes as mentioned in #53 .

sort


dashboard/views/users.ejs
```diff
@@ -137,10 +144,9 @@ function show_class_fields() {
...
         $('.classroom_input').addClass('hide');
         $('#user-form [name="classid"]').val('').trigger('change');
-
     }else{
         $('.classroom_input').removeClass('hide');
     }
 }
 show_class_fields('<% if(typeof query == "object"){ %><%= query.role || "student" %><% } %>')
-</script>
+</script>
```
tarunsinghal92: new line here!

ashish0910:
@tarunsinghal92 , @llaske changes done .

llaske:
It's better. 
Regarding the behavior we expect to be able to reverse the sort order by reclicking on the column header. It's not the case today.
Regarding the UI, the arrows should be less visible (not on a background) and only one arrow should be visible at the same time. Something like that: https://demos.telerik.com/kendo-ui/grid/index

ashish0910:
@llaske this should work now 

ezgif com-video-to-gif (1)

Similarly added for ->
user - name , last seen
activity - name , activity id , version 
journal - timestamp , size
classrooms - name , last seen

llaske:
Nice. It's exactly the UI expected. 
BTW your sort works only on displayed result not on the whole content. So if you've got 100 users an the last one is "aaaa", it will not appear in first page.
Plus sort must not be done on activities because the order define order in the home view.

ashish0910:
@llaske I don't think it is the case .
As you can see 'aaaaa' student was created last and when we sort +name 'aaaaa' comes at first of first page and 'z' comes at last of last page and when we sort -name z comes at first of first page and 'aaaaa' comes at last of last page and all other students are also sorted perfectly . The total students are sorted first and are then divided for pagination . Actually earlier the arrows were not shown on paginated pages that might have caused the confusion but now it is added .
It works perfectly now .

ezgif com-video-to-gif

Anshuman71:
@ashish0910 @llaske @tarunsinghal92  should we add sort by students count in classrooms?
I think the admin sometime may be interested in that also.

ashish0910:
@Anshuman71 sure we can have that

ashish0910:
@Anshuman71 the current API response does not have a field student_count instead it has an array of students . Sorting based on student count will require adding that field which is not feasible right now .

Anshuman71:
> @Anshuman71 the current API response does not have a field student_count instead it has an array of students. Sorting based on student count will require adding that field which is not feasible right now .

Ok, thanks for follow up

ashish0910:
@llaske it should work perfectly now

ashish0910:
@llaske can you please review this pr

ashish0910:
@llaske @tarunsinghal92 can you please review this pr

ashish0910:
@llaske @tarunsinghal92 can you please review this pr

llaske:
Something goes wrong. The username column header is missing

image

And there is an error in the console:

image

ashish0910:
@llaske fixed the problem . That probably came while solving merge conflicts .

ezgif com-video-to-gif

llaske:
Sorry, still not work on my side. Clicking on sort by Username don't give the result expected.

ashish0910:
@llaske that is happening because of mix of lowercase and uppercase . The uppercase comes first in case of ascending order . 
As of now mongodb does not support case insensitive sorting
[ref1](https://stackoverflow.com/questions/22931177/case-insensitive-sorting-in-mongodb)
[ref2](https://forums.meteor.com/t/how-to-perform-case-insensitive-sorting/1103)
[ref3](https://github.com/doctrine/mongodb-odm/issues/1237)
In order to handle this situation we need to add another field in database which will store data in lower case but I don't think it will be a good option . There is another alternative that we can convert the user automatically in lower case while adding user . As of now we are treating users as case insensitive while adding ie 'Ashish' and 'ashish' will be same but in db raw value is stored which I think is not good as it can cause inconsistency in the db . Also we can modify API in a way to convert data to lower case while fetching the users .
@llaske @tarunsinghal92 please let know what should be the best way to overcome this situation ?

PR #80, opened 2019-03-10
[WIP] Installing Gulp

ashish0910:
We can use gulp to minify and manage public resources as mentioned [here](https://github.com/sugarlabs/GSoC/blob/master/Ideas-2019.md) . 
This initialises gulp for the project . Shall we also use gulp it for sass and scss ?
Also I need some recommendation about the resultant project structure produced by gulp .
@tarunsinghal92 @llaske

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,6 +1,6 @@
-var gulp = require('gulp'),
-var imagemin = require('gulp-imagemin'),
-var uglify = require('gulp-uglify');
+const gulp = require('gulp');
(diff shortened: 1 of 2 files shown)


llaske:
Sugarizer client code is minified using [Grunt](https://github.com/llaske/sugarizer/blob/dev/Gruntfile.js) instead of Gulp.
Not sure which one is better.
BTW adding this feature is not just adding dependencies, it will require rethink the build and deploy process of the dashboard so I'm more in favor to add an issue and let this task for later.

PR #83, opened 2019-03-10
Fix server hang/crash on admin deletion

### LINKED ISSUE #82: Server hangs/crashes on when an admin delete its own account
When the currently logged in admin user deletes their account, the server hangs and the page is stuck on reload until it crashes.

**Problem:**

admin logout

**Fix: #83**

Disallow currently logged in admin from deleting their own account and display an error message.

delete admin

ahmedkrmn:
Fixes #82
Disallow currently logged in admin from deleting their own account and display an error message. Details on the issue page. 
This crash is caused by inconsistencies between the cookies and the stored user session when the logged in admin deletes his/her account.


llaske:
Thanks. Nice fix.

PR #89, opened 2019-03-11
UX Improvement. Hide non editable empty fields.

NikhilM98:
While creating Classroom/Users. Empty fields are displayed for `Created Time` and `Last Seen`. They look like they have to be filled. It creates confusion for the users whether to fill them or not.

Current:
Sugarizer Dashboard (2)

Updated:
- Users
Sugarizer Dashboard

- Classroom
Sugarizer Dashboard (1)

The `Created Time` and `Last Seen` fields should appear while editing. Not while creating.

Also, I changed `Last Seen` in the classroom to `Last Updated` as the timestamp indicated the last time classroom was updated.


ashish0910:
@NikhilM98 rather than removing these fields we can show user current timestamp in both these fields .

NikhilM98:
@ashish0910 timestamp is added in the backend. If we show current timestamp in the field and the user takes 10 minutes to add a classroom then he won't we adding the timestamp he's expected.

llaske:
Good idea.

PR #92, opened 2019-03-12
Classroom user count fix

NikhilM98:
The student count of a classroom is not updated when we delete a user.

Current Behavior:
Sugarizer Dashboard (4)

Expected Behavior:
Sugarizer Dashboard (3)

API Test:
Screenshot from 2019-03-12 13-27-34

### DIFF
--- a/api/controller/users.js
+++ b/api/controller/users.js
@@ -7,10 +7,13 @@ var mongo = require('mongodb'),
 var db;
 
 var usersCollection;
+var classroomsCollection;
+
 
 // Init database
 exports.init = function(settings, callback) {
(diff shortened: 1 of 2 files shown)


llaske:
Hmmm. Well done.
By the way, I think we should add a new unit test in classroom regarding this case.
Could you add it to your PR?

NikhilM98:
@llaske I've added tests for checking the user count of the classroom.
- When the classroom is created, it's created with two students.
- On deleting a student, tests for the students in a classroom.

llaske:
Good job.

PR #93, opened 2019-03-12
Fix the ordering of activities while sorting

NikhilM98:
While sorting the activities, the numbering of the activities appears ambiguous on picking up the element.
On dragging an item, it appears as if there is a vacancy of two elements instead of one.

Current Behavior:
Sugarizer Dashboard (6)

Expected Behavior:
Sugarizer Dashboard (5)

Solution: The wrong ordering is due to the hidden `<li>` placeholder element. I used `<div>` for the placeholder instead so it's not causing ambiguities in the ordering of the list.


llaske:
Okay, it's a minor issue. But it's a good point.

PR #94, opened 2019-03-13
size column in journal view

ashish0910:
Earlier wrong field was used for showing size in journal view as mentioned by @llaske [here](https://github.com/llaske/sugarizer-server/pull/67#issuecomment-472199663)

Screenshot 2019-03-13 at 9 10 28 AM

### DIFF
--- a/dashboard/views/journal.ejs
+++ b/dashboard/views/journal.ejs
@@ -99,7 +99,7 @@
 																	</script>
 																	<td><%= entries.entries[i].metadata.title %></td>
 																	<td><%= moment(entries.entries[i].metadata.timestamp).calendar() %></td>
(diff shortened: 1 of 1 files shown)


llaske:
Look like it don't works for all size. May be it's related to column size ?

image

Plus, miss a white space before "Bytes".

ashish0910:
@llaske This should work now 

Screenshot 2019-03-13 at 1 37 16 PM

llaske:
Nice. Could you localized strings like it was in [Sugarizer](https://github.com/llaske/sugarizer/blob/dev/lib/util.js#L543).

ashish0910:
@llaske done in #104

PR #96, opened 2019-03-13
Fix Statistics page showing all users as inactive

### LINKED ISSUE #95: Statistics page shows all users as inactive
The **How many users actives during last month?** chart on the _statistics_ page shows that all users were inactive during last months even though they were!

inactive

Fixed here: #97

ahmedkrmn:
Fixes #95 

- All users were active during last month:
active 1

- Three users were active and two users weren't:
active 2


ahmedkrmn:
Because of the redundant use of  `unshift()`, I've created a new PR here #97.

PR #97, opened 2019-03-13
Fix Statistics page showing all users as inactive

### LINKED ISSUE #95: Statistics page shows all users as inactive

ahmedkrmn:
Fixes #95 

- All users were active during last month:
active 1

- Three users were active and two users weren't:
active 2

### DIFF
--- a/dashboard/controller/stats.js
+++ b/dashboard/controller/stats.js
@@ -182,15 +182,13 @@ function getHowUsersAreActive(req, res) {
 	// get data
 	getUsers(req, res, { role: 'student' }, function(body) {
 		total = body.total;
-		data.data.push(body.total);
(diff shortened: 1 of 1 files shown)


llaske:
Very nice. Thanks a lot.
Could you check if there is no error on other stats?

ahmedkrmn:
@llaske just went through the code of other statistic graphs. Everything is working as expected 👍

ahmedkrmn:
@llaske #119

Anshuman71:
> @llaske #119

thank you @ahmedkrmn for referencing the issue I have made new PR as we discussed

PR #100, opened 2019-03-13
Show an alert when User name or Class name is empty

ahmedkrmn:
Current behavior is that an alert is shown in case of an empty user name or class name but only when creating a new one. 
Nothing is shown in case of editing an existing one.

empty class

---

**Fix:** Show an alert when user name OR class name is empty

empty class dddaa

_Note: In case of creation of an unnamed class, the error shown says: "Username is not valid" so I updated it to "Classname is not valid."_

---

**Further modifications:** 3055afb
In order to keep the student names entered when creating a new class, validation should be on the client side instead. 
Retrieving the entered values back through the API as in PR #99 is not possible in this case.

empty class dddaadddsds

empty class2aaa

---

**Request changes:** cd5851a

aaa


llaske:
We must forbid to save change when the username or classname is empty.
Displaying a popup is too late.

ahmedkrmn:
@llaske Added user name validation to the client side so updating is forbidden.

dashboard/views/addEditClassroom.ejs
```diff
@@ -14,7 +14,7 @@
                             <div class="col-md-12">
                                 <div class="form-group form-black label-floating">
                                     <label class="control-label" data-l10n-id="name"></label>
-                                    <input class="form-control" name="name" value="<% if(typeof classroom == 'object'){ %><%= classroom.name %><% } %>" type="text">
+                                    <input class="form-control" name="name" value="<% if(typeof classroom == 'object'){ %><%= classroom.name %><% } %>" type="text" required>
```
llaske: I'm not in favor of using "required". It's not consistent with other forms plus it don't use the same localization than the server. You could just let the server do the work.

dashboard/views/addEditUser.ejs
```diff
@@ -14,7 +14,7 @@
                             <div class="col-md-12">
                                 <div class="form-group form-black label-floating">
                                     <label class="control-label" data-l10n-id="name"></label>
-                                    <input class="form-control" name="name" value="<% if(typeof user == 'object'){ %><%= user.name %><% } %>" type="text">
+                                    <input class="form-control" name="name" value="<% if(typeof user == 'object'){ %><%= user.name %><% } %>" type="text" required>
```
llaske: I'm not in favor of using "required". It's not consistent with other forms plus it don't use the same localization than the server. You could just let the server do the work.

ahmedkrmn: I believe there is no way we can have validation on the server without showing the popup. The page needs to be reloaded and thus the popup shows.

PR #101, opened 2019-03-13
Fixes breaking tests. Handling admin deletion 

NikhilM98:
The tests were failing after PR #83 

Moved the logic to check that if the admin is deleting himself to `dashboard/controller`

Failing tests:
Screenshot from 2019-03-13 17-03-13

After fix:
Screenshot from 2019-03-13 17-03-30

The functionality introduced by #83 is also handled and is working as expected.

### DIFF
--- a/api/controller/users.js
+++ b/api/controller/users.js
@@ -627,14 +627,6 @@ exports.removeUser = function(req, res) {
 			});
 			return;
 		}
-	} else{
-		if (req.user._id == req.params.uid) {
-			res.status(400).send({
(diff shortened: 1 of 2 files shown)


ahmedkrmn:
Yeah, I see. This occurred because in case of admin deletion, I sent the response with a status code of 400. The test was expecting a status code of 200.
I really like the way you did it here!

llaske:
Very good job @NikhilM98. 
@ahmedkrmn, please launching unit test before sending a PR is good practice.

PR #102, opened 2019-03-13
[UX Improvement] Create another user or classroom without returning to the main page

ahmedkrmn:
If the admin wants to create another user or classroom, he/she has to click on the **ADD CLASSROOM** or **ADD USER** button again from the main page.
This would by a length process if the admin wants to create multiple users or classrooms at a time.

create1
 
---

**Fix:** We can have two buttons for saving. One for exiting upon creation and another one for adding another user/class.

empty class dddaadddsdsa

In case of editing a classroom or a user, the **SAVE AND ADD ANOTHER** button is not displayed.


llaske:
We plan to add an import action to create multiple classrooms/users using a CSV file so this feature is not need.

ahmedkrmn:
@llaske yeah that's definitely a better idea.

PR #104, opened 2019-03-14
localisation added in size column

ashish0910:
localized strings as mentioned by @llaske in #94 

Screenshot 2019-03-14 at 7 20 11 PM

Screenshot 2019-03-14 at 7 20 21 PM


llaske:
Thanks. Nice.

PR #105, opened 2019-03-14
[WIP] Script to seed test data

NikhilM98:
I've created a script to seed the database with Fake Students, Fake Admins, and Fake Classrooms with Fake Students as discussed with @tarunsinghal92 

The script can be run by using `npm run seed:db`
- Creates a Master Admin and displays its username and password. Can be used by the user to log in.
- Creates Fake Students and Fake Admins.
- Creates Fake Classrooms with a random number of Fake Students.

I've some doubts about creating Fake Journal mapped to Fake Students.
How much fake the Fake Journals have to be?
- Should fake journals entries have to be related to all available activities and have accurate data corresponding to those activities. 
or
- Should fake journals entries can be related to 5-6 activities and have corresponding data. Students can randomly have any number of journal entries.

Screenshot:
Screenshot from 2019-03-14 20-20-28

P.S. Currently it's hardcoded to seed 5 Students, 5 Admins, and 5 Classrooms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
 		"test": "mocha api/test/",
 		"docs": "apidoc -i api/controller -i dashboard/helper -o docs/www/",
 		"start": "node sugarizer.js",
-		"start:watch": "nodemon sugarizer.js"
+		"start:watch": "nodemon sugarizer.js",
+		"seed:db": "node seed.js"
 	}
 }
--- a/seed.js
+++ b/seed.js
@@ -1,0 +1,221 @@
+// Import request library to make requests over the network
+var request = require('request');
+
+// Import common helper
+var common = require('./dashboard/helper/common');
+
+// Set the node enviornment to test
+process.env.NODE_ENV = 'test';
+
+// Import sugarizer server
+require('./sugarizer');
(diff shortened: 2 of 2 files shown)


llaske:
Not sure to understand why we need this script. Is it for testing?

llaske:
@NikhilM98 we need a script to init the database for our current deployment with something similar to your script but with a CSV file as input. Could you adapt your script ?

The idea will be to use an input CSV file with columns:

`Name,[Type],[Language],[Color],[Password],[Classroom]`

Everything except Name is optional and will be set to default if not provided: default type is student, default language is en, default color is random, default password is random, no classroom by default.

Rules will be for each line:

- Create a new user with properties,
- Add the user in the matching classroom,
- If an user with this name already exist, write a warning and ignore the line,
- If a classroom with this name don't exist, create it.

The script should produce a CSV file with all properties for created users (to have a way to retrieve generated password).

NikhilM98:
@llaske Okay. I'll do it.

llaske:
No longer need since https://github.com/llaske/sugarizer-server/pull/168

PR #107, opened 2019-03-14
Search for stroke and fill fixed

NikhilM98:
The search for stroke and fill color was not working while creating user/classroom.

Current Behavior:
Screenshot from 2019-03-14 22-32-18

Expected Behavior:
Screenshot from 2019-03-14 22-34-06


llaske:
Hmmm. Not sure lot of people will search for colors but because the search field is there, it's better if it works.

PR #110, opened 2019-03-16
Updated translations

NikhilM98:
- Translated stats labels which were not previously translated.
Screenshot from 2019-03-16 16-34-15
Screenshot from 2019-03-16 16-28-51

- Corrected and added some translations.


llaske:
Ok. Nice.

PR #114, opened 2019-03-17
fix #112

### LINKED ISSUE #112: No message when searched activity is not found
When we miss a search in activities we are not shown any message regarding the `no match found` which seems odd and is not quite a good user experience.
I have fixed this in #125

Anshuman71:
Add error label when no match found
Screenshot from 2019-03-17 10-58-33

1. Show/hide label accordingly
2. Search using 'Enter' on the keyboard
3. Toggle label on empty input

I hope this improves the user experience


NikhilM98:
@Anshuman71 can you keep the file changes minimal so that it's easier to review what's changed in the code.

Anshuman71:
@NikhilM98 My code editor formatted the code sorry for that. I will make sure it doesn't happen next time

PR #115, opened 2019-03-17
Delete private journal on delete user

### LINKED ISSUE #109: Delete all user data when the user is deleted
When a user is deleted, his journals and stats are not removed from the database.
I was wondering whether should we remove the user data when the user is deleted. The data can be useful for analytics when we expand sugarizer-server dashboard. But storing someone's data without their consent is also not good.

NikhilM98:
Fixes #109 
Delete private journal on delete user

Tests passed:
Screenshot from 2019-03-17 14-11-44

I've checked the journals collection. It's not storing extra data after test completion. (Previously it was storing private journals unnecessarily) 

I had to modify the test for users. It was deleting journal after deleting the user. But since the journal is deleted with the user, the test was failing so I had to make appropriate changes.

### DIFF
--- a/api/controller/users.js
+++ b/api/controller/users.js
@@ -14,6 +14,7 @@ var classroomsCollection;
 exports.init = function(settings, callback) {
 	usersCollection = settings.collections.users;
 	classroomsCollection = settings.collections.classrooms;
+	journalCollection = settings.collections.journal;
 	var client = new mongo.MongoClient(
 		'mongodb://'+settings.database.server+':'+settings.database.port+'/'+settings.database.name,
 		{auto_reconnect: false, w:1, useNewUrlParser: true});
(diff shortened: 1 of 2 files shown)


tarunsinghal92:
@llaske 

Do we keep the private journals even after deleting the user?

llaske:
@tarunsinghal92 no we should not.

NikhilM98:
It is also causing inconsistency with the number of journals displayed on the dashboard.
Screenshot from 2019-03-26 17-19-02

You can see that on the first chart, the total entries are 20 but it's counting the total number of journals as 25 on the third card due to the undeleted journals.

ashish0910:
@NikhilM98 first chart is for the frequency of number of contributions of top contributors not the count of journals

NikhilM98:
> @NikhilM98 first chart is for the frequency of number of contributions of top contributors not the count of total journals

@ashish0910, if you analyze the chart, you'll notice that in this case, the summation of the total number of entries of the top contributors is the count of the total journals entries.

ashish0910:
@NikhilM98 yes it is because only 2 users had contributions and remaining did not . If there were 100 users then top contributors chart will show for only top few and their summation will be less than total contributions of 100 users .

NikhilM98:
@ashish0910 You're not understanding the point.
Since you agree here:
> @NikhilM98 yes it is because only 2 users had contributions and remaining did not.

I suppose you agree that there are 20 total journal entries.
Now if you look at the third card, it'll show you 25 total entries.
It's because the entries were not deleted when the user was deleted.

PR #120, opened 2019-03-18
Fix #119

### LINKED ISSUE #119: `How users are launching activities` not showing any data
Is this only me or the statistics page is not showing a chart for `How users are launching activities` when there is indeed an entry in the database collection 'stats'
Screenshot from 2019-03-18 09-46-39

I am on this issue and have found the bug.
Please let me know if this is an issue indeed

Anshuman71:
Change the assignment of  0 with 1 when we add it for 1st time to object
Before
Screenshot from 2019-03-18 09-46-39
After
Screenshot from 2019-03-18 09-59-13


ahmedkrmn:
As for your fix a5eabc3e5a7acc5341a069207860ac1eebb0e0e3, yeah that's the way it should be done 👍

Anshuman71:
> As for your fix [a5eabc3](https://github.com/llaske/sugarizer-server/commit/a5eabc3e5a7acc5341a069207860ac1eebb0e0e3), yeah that's the way it should be done

thanks, I will remove the other two issue and will make a new pr and reference this one there

Anshuman71:
> As for your fix [a5eabc3](https://github.com/llaske/sugarizer-server/commit/a5eabc3e5a7acc5341a069207860ac1eebb0e0e3), yeah that's the way it should be done

thanks, I will remove the other two issue and will make a new pr and reference this one there

Issue #111, opened 2019-03-17
Search for users in journal screen is limited


ashish0910:
On `dashboard/journal?lang=en` route when you click on select user option it shows only 10 most recent users and not others and when you search the user which is left then it shows no result found . Hence we are not able to access journal of all users . 

as you can see user 'b' exists 

![Screenshot 2019-03-17 at 8 37 09 AM](https://user-images.githubusercontent.com/30404493/54484593-f6d22080-488f-11e9-8a79-4195d5a32a40.png)

On journal page on searching 'b' , it shows no result found

![Screenshot 2019-03-17 at 8 37 32 AM](https://user-images.githubusercontent.com/30404493/54484601-19fcd000-4890-11e9-868d-f9c619b167e7.png)

I have started working on this issue

llaske:
You're right. It could be improve.

ashish0910:
@llaske that did not work in my case 

![journal](https://user-images.githubusercontent.com/30404493/55642871-4f417180-57ef-11e9-8b57-a43e78d64253.gif)

Also there is another issue in it , when you go to users page and click on show journal icon for a user who is not in the list of the most ten recent users then it shows the name of most recent user but shows the desired data that is the of the user selected .

![journal2](https://user-images.githubusercontent.com/30404493/55643237-4dc47900-57f0-11e9-8701-62cfe51bfccf.gif)

NikhilM98:
Fixed in #213

PR #122, opened 2019-03-18
fix issue #111

ashish0910:
Now journal will show all users in select tag .

### DIFF
--- a/dashboard/controller/journal.js
+++ b/dashboard/controller/journal.js
@@ -163,7 +163,8 @@ function getUsers(req, res, callback) {
 		method: 'GET',
 		qs: {
 			sort: '-timestamp',
-			role: 'student'
+			role: 'student',
+			limit: 1000
 		},
 		uri: common.getAPIUrl(req) + 'api/v1/users'
 	}, function(error, response, body) {


ashish0910:
@NikhilM98 yes but there is a drawback there that if you want to search a user which is at 40th position then it might create a problem for the admin as he will not be able to find him in the list

NikhilM98:
@ashish0910 I'm saying that Select2 has support for [Ajax requests](https://select2.org/advanced/default-adapters/ajax). With some modification in the backend, we'll be able to search for the users on the server side. Transferring data of thousands of users in a single request doesn't seem efficient.

ashish0910:
@NikhilM98 yes that will be good . I will change it accordingly .

ashish0910:
@llaske what are your views on this as adding a realtime search in select through AJAX might put some load as for example when admin types 'a' the query will start in background and start searching all the students with letter 'a' which will be in huge number and then when he press 's' then query will search for all users with 'as' in name which will again be huge number and process will go on like this . This is not looking very efficient .
I propose a search for name of the user using an input field as in user view and if the user do not exist a message will be shown saying no user found .

NikhilM98:
@ashish0910 The search can be optimized.
- Set a limit for the number of users selected in one query.
- In the Select2 library, we're using, there's an option to set delay. You can set the delay to 250ms so as to avoid unnecessary queries.
- Optional: Set a minimum number of terms after which you start searching

ashish0910:
@tarunsinghal92 can we have search similar to that in user views and show results if the user exists otherwise show user doesn't exist ?

PR #123, opened 2019-03-18
Fix chart not showing any data for single entry

Anshuman71:
Change the assignment of  0 with 1 when we add it for the 1st time to object #119 #120

### DIFF
--- a/dashboard/controller/stats.js
+++ b/dashboard/controller/stats.js
@@ -60,7 +60,7 @@ function getHowUserLaunchActivity(req, res) {
 			var d = {}
 			for (var i = 0; i < data.length; i++) {
 				if (!d.hasOwnProperty(data[i].event_object)) {
-					d[data[i].event_object] = 0
+					d[data[i].event_object] = 1
 				} else {
 					d[data[i].event_object]++
 				}


dashboard/controller/stats.js
```diff
@@ -60,7 +60,7 @@ function getHowUserLaunchActivity(req, res) {
 			var d = {}
 			for (var i = 0; i < data.length; i++) {
 				if (!d.hasOwnProperty(data[i].event_object)) {
-					d[data[i].event_object] = 0
+					d[data[i].event_object] = 1
```
NikhilM98: If you set it to 1, then it'll generate the chart, but it'll show incorrect data. It'll show some has already launched one activity from the `List View` and one from the `Journal View`.

ahmedkrmn: I've doubled checked and it's working correctly for that case. Can you please elaborate, maybe some screenshots?

NikhilM98:
> If you set it to 1, then it'll generate the chart, but it'll show incorrect data.
> When you hover on the chart, it'll show that some user has already launched one activity from the `List View` and one from the `Journal View`.

@Anshuman71 @ahmedkrmn I checked the data again. I admit my last analysis was incorrect.
This patch will not only solve the issue with the rendering of the chart for single data, but it'll also fix the count which was initially started from the wrong number.

Anshuman71:
> > If you set it to 1, then it'll generate the chart, but it'll show incorrect data.
> > When you hover on the chart, it'll show that some user has already launched one activity from the `List View` and one from the `Journal View`.
> 
> @Anshuman71 @ahmedkrmn I checked the data again. I admit my last analysis was incorrect.
> This patch will not only solve the issue with the rendering of the chart for single data, but it'll also fix the count which was initially started from the wrong number.

Thank you @NikhilM98 for the feedback and support.

Anshuman71:
@tarunsinghal92  please review this change.

llaske:
@Anshuman71 you're right. The count was wrong. Thanks.

PR #85, opened 2019-03-11
Go to User journal / Launch activity when clicked on chart

ahmedkrmn:
Fixes #84
Go to the User journal page OR Launch activity page when admin clicks on the chart.

**Top Contributors chart:**

bar chart

**Top Activities chart:**

pie chart

PR #128, opened 2019-03-19
Fix console error on bar chart onClick

NikhilM98:
Steps to reproduce the error:
1. Open dashboard. Navigate to statistics page.
2. Keep the developer console open and click on "How often users are changing their profile settings?" chart.
You'll get an error like this:
Screenshot from 2019-03-19 13-20-18

Cause of error: In PR #85 an event listener was added to all the charts. It should be added to only a couple of charts. The error was popping because of the event listener being attached to the wrong chart and the function breaking down because of undefined variables.

In this PR:
- Instead of attaching the event listener to all the charts. It's attached to only selected charts thus eliminating the error and the improving efficiency.
- ES6 template literals were removed due to lack of support in the older browsers version.

### DIFF
--- a/dashboard/public/js/main.js
+++ b/dashboard/public/js/main.js
@@ -299,7 +299,7 @@ function createGraph(type, element, route) {
 				var html = '<div class="text-center">\
 											<i class="material-icons dp96 text-muted">info_outline</i>\
 											<p data-l10n-id="noGraphDataText">' + document.webL10n.get('noGraphDataText') + '</p>\
-										</div>'
+										</div>';
 				$("#" + response.element).replaceWith(html);
 			} else {
 				var ctx = document.getElementById(response.element).getContext('2d');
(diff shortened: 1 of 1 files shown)


dashboard/public/js/main.js
```diff
@@ -308,19 +308,28 @@ function createGraph(type, element, route) {
 					data: response.data,
 					options: (response.options ? response.options : {})
 				});
-				myChart.options.onClick = function(e){
-					var activePoints = myChart.getElementsAtEvent(e);
-					// Avoid console erros when clicking on any white space in the chart
-					var index = activePoints.length ? activePoints[0]._index : -1;
-					if (index > -1 && response.graph === 'bar'){
```
ahmedkrmn: Way easier if you could've just changed the condition to:
```suggestion
					if (index > -1 && type === 'top-contributor')
```

dashboard/public/js/main.js
```diff
@@ -308,19 +308,28 @@ function createGraph(type, element, route) {
...
-						window.location.href = `/dashboard/journal/${response.journalIDs[index]}?uid=${response.userIDs[index]}&type=private`;
-					}else if(index > -1 && response.graph === 'doughnut'){
```
ahmedkrmn: And this condition to:
```suggestion
					}else if(index > -1 && type === 'top-activities')
```

NikhilM98:
> Way easier if you could've just changed the condition to:
> 
> Suggested change 
> 					if (index > -1 && response.graph === 'bar'){
> 					if (index > -1 && type === 'top-contributor')


@ahmedkrmn that won't be an efficient way.
As I mentioned:
> In this PR:
> 
> * Instead of attaching the event listener to all the charts. It's attached to only selected charts thus eliminating the error and the improving efficiency.
> * ES6 template literals were removed due to lack of support in the older browsers version.

If I do so something like this then:
1. It would be considered unnecessarily adding event listeners to all the charts.
2. We're not using any ES6 transpilers currently, so it may happen that slightly older browsers won't be able to recognize the ES6 template literals.

ahmedkrmn:
Don't think that those event listeners would have any noticeable performance impact, but agree with you on using ES5 instead 6. I'll make sure to use ES5 in my code henceforth.

Issue #131, opened 2019-03-20
private and shared select tag of journal translation issue


ashish0910:
After #124 a bug has arrived in translation of select tag of shared and private journal . After changing the language for couple of times this problem arises .

![Screenshot 2019-03-20 at 9 36 08 AM](https://user-images.githubusercontent.com/30404493/54658873-8fb5a580-4af5-11e9-95a3-be31f3c154ee.png)

![Screenshot 2019-03-20 at 9 36 01 AM](https://user-images.githubusercontent.com/30404493/54658883-96441d00-4af5-11e9-90f4-5707bc6a3152.png)

I have started working on this issue

ashish0910:
fixed in #132

PR #121, opened 2019-03-18
Dropdown language change fix

NikhilM98:
Following the discussion on PR #43 
- Server translations are reloaded on locale change.
- Links on the dashboard were missing locale parameter.
- The cursor on the sidebar links was not appearing as a pointer.

PR #132, opened 2019-03-20
#131 fixed 

ashish0910:
This pr fixes issue created in #124 as mentioned in #131 

Screenshot 2019-03-20 at 10 18 14 AM

Screenshot 2019-03-20 at 10 20 17 AM

### DIFF
--- a/dashboard/views/journal.ejs
+++ b/dashboard/views/journal.ejs
@@ -40,7 +40,13 @@
 								<script>
 								window.addEventListener('localized', function() {  // HACK: Force localization of selected field
 									var selection = document.getElementsByClassName("select2-selection__rendered");
-									if (selection && selection[1] && typeof selection[1].innerHTML == "string") selection[1].innerHTML = document.webL10n.get(selection[1].innerHTML.toLowerCase());
+									if (selection && selection[1] && typeof selection[1].innerHTML == "string") {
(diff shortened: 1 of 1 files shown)


NikhilM98:
Maybe I made changes following the PR [#121](https://github.com/llaske/sugarizer-server/pull/121). That's why this issue was not appearing for me.

This way of selecting the selected value of the select field is clearly more accurate. I tried to work around with the previously used method but seems to have created an issue.

dashboard/views/journal.ejs
```diff
@@ -40,7 +40,13 @@
 								<script>
 								window.addEventListener('localized', function() {  // HACK: Force localization of selected field
 									var selection = document.getElementsByClassName("select2-selection__rendered");
-									if (selection && selection[1] && typeof selection[1].innerHTML == "string") selection[1].innerHTML = document.webL10n.get(selection[1].innerHTML.toLowerCase());
+									if (selection && selection[1] && typeof selection[1].innerHTML == "string") {
+										var journal_select_type = document.getElementById("journal-type-select2");
+										var val = journal_select_type.options[journal_select_type.selectedIndex].value;
+										if (selection && selection[1]) {
```
NikhilM98: There's no need to use the null checks for `selection && selection[1]` here. It's already checked in the previous `if` loop.

llaske:
Nice.

PR #134, opened 2019-03-20
Re-init l10n in stats controller

NikhilM98:
The l10n instance was not initiated on the stats controller.
If a user switched language and moved to stats page, the language of the graphs was not changed.

Current Behavior:
Sugarizer Dashboard (11)

Expected Behavior:
Sugarizer Dashboard (12)

### DIFF
--- a/dashboard/controller/stats.js
+++ b/dashboard/controller/stats.js
@@ -13,8 +13,9 @@ exports.init = function(settings) {
 // main landing page
 exports.index = function(req, res) {
 
-	// reinit momemt with locale
+	// reinit l10n and momemt with locale
 	if (req.query && req.query.lang) {
+		common.l10n.setLanguage(req.query.lang);
 		moment.locale(req.query.lang);
 	}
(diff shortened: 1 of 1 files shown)


NikhilM98:
> It fix the issue when the user move to another page but not for charts on the current page.
> i.e. if I'm in french and I'm switching in english, tooltip stay in french until I move to another page.

@llaske We need to reload all the charts again in order to fix the issue you mentioned.
PR #121 can handles that case.

NikhilM98:
> It fix the issue when the user move to another page but not for charts on the current page.
> i.e. if I'm in french and I'm switching in english, tooltip stay in french until I move to another page.

@llaske We need to reload all the charts again in order to fix the issue you mentioned.
There's an existing open PR that handles the case you mentioned. Can you please review #121.

llaske:
Ok I see. I've merged https://github.com/llaske/sugarizer-server/pull/121 too.

PR #135, opened 2019-03-20
Fixed placeholder translation in addEditClassroom

NikhilM98:
The placeholder in the addEditClassroom view for `Search Students` field was not being translated due to a typo and we were also getting warnings in the console.
Screenshot from 2019-03-20 14-53-59

Fixed:
Screenshot from 2019-03-20 14-48-00


llaske:
Nice.

PR #137, opened 2019-03-21
 Display name in notification after create/update/delete classroom/user operations

NikhilM98:
Following the conversation in #103, I've modified the popup to display the name of the user/classroom after the create/update/delete operations.

Test Screenshots:
Screenshot from 2019-03-21 18-18-32
Screenshot from 2019-03-21 18-18-56
Screenshot from 2019-03-21 18-19-09
Screenshot from 2019-03-21 18-20-24
Screenshot from 2019-03-21 18-20-32
Screenshot from 2019-03-21 18-20-38

PR #103, opened 2019-03-13
UX Improvement. List admins after create/update/delete admin

NikhilM98:
I was facing this issue while testing PR #101.

After creating/editing/deleting an admin, the app returns me to users view listing students instead of admins. It's really frustrating to delete/edit multiple admins and It's difficult to keep track of your last action.

Current behavior:
Sugarizer Dashboard (8)

Expected behavior:
Sugarizer Dashboard (7)
(When user creates/updates/deletes admin, he gets the list of admins)

PR #138, opened 2019-03-21
Display name of the activity while deleting journal entry

NikhilM98:
Following the discussion on PR #103, I've made changes in the popup for user/class created/updated/deleted in PR #137

I found a possibility of similar UX improvement in the Journals view, where: 
- Name of the activity is not specified in the confirmation.
- After deleting an entry, the popup does not specify which entry was deleted.

Current:
current

Improvement:
correct


llaske:
Nice. Thanks.

PR #140, opened 2019-03-23
Fix API inconsistency for asynchronous calls

### LINKED ISSUE #139: Inconsistency in the function execution
In the `process-results` function in `api/controller/activities`.

The variables for the array `activities2` and the flag `isValid` are defined as global variables.
Since JS is asynchronous in nature and so is Node, this can lead to inconsistency in the array and flags in the case the function is executed at the same time by two or more users (If the flag is raised for one user, it will also be raised for the second user. It's also possible that the array will end up returning the same value for parallel executions)

This bug has a slight possibility of being detected in case of a few users, but for the long term, it's better if it's resolved.

https://github.com/llaske/sugarizer-server/blob/2cf2d524055323ded375a310ba4259aadde5ee9f/api/controller/activities.js#L230-L237

https://github.com/llaske/sugarizer-server/blob/2cf2d524055323ded375a310ba4259aadde5ee9f/api/controller/activities.js#L252-L256

For now, it seems like a minor issue. Should I fix this?

NikhilM98:
Fixes #139 

Following issue #139, I've identified and fixed potential global vars in API which can cause inconsistency during parallel function calls.

### DIFF
--- a/api/controller/activities.js
+++ b/api/controller/activities.js
@@ -195,7 +195,7 @@ exports.findAll = function(req, res) {
 exports.findById = function(req, res) {
 
 	//process results based on filters and fields
-	data = process_results(req, activities);
+	var data = process_results(req, activities);
 
 	//find by id
 	var id = req.params.id;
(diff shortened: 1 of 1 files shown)


llaske:
You're right. Could happen.

PR #141, opened 2019-03-23
Show view journal button only for students in users view

NikhilM98:
In general case, there seems no point of showing view journal button for admin and teachers on the users view. It should be displayed only for the student accounts.

Student:
Screenshot from 2019-03-23 20-19-12

Admin/Teacher:
Screenshot from 2019-03-23 20-19-04


llaske:
Make sense.

PR #143, opened 2019-03-24
Fix classroom for similar students

NikhilM98:
While using filters in the `Users View`. If Class A and Class B have the same students. If we use the filter for Class B, then it displays the name of Class B in the select field.

Sugarizer Dashboard (14)

Fixed:
Sugarizer Dashboard (15)


llaske:
Waow, very complex issue to find. Thanks.

PR #145, opened 2019-03-25
Added localisation for the language of user

ashish0910:
Currently language of user is not translated and is shown in english only . This PR added translation for the language of user .

Current behaviour:

Screenshot 2019-03-25 at 7 23 26 PM

Expected behaviour:

Screenshot 2019-03-25 at 7 22 24 PM

PR #146, opened 2019-03-25
localisation for translating language

ashish0910:
Currently the select box shows the available languages in English only . This PR added localisation for that select box .
Current Behaviour:
Screenshot 2019-03-25 at 7 55 36 PM

Expected Behaviour:
Screenshot 2019-03-25 at 7 51 58 PM


ashish0910:
@NikhilM98 Thanks for your suggestion , as long as project is improved I don't think there should be any problem in this also I did not see any pr for this change in this project so I opened it . 
Regarding the taking local language of browser , I will do the changes accordingly .

NikhilM98:
@ashish0910 
- Use of `english-nav`, `hindi-nav`, `spanish-nav` and `french-nav` in locales looks redundant when compared to #145

- I don't understand what you mean by:
> Regarding the taking local language of browser 
 
But implementing double translation while fixing English names doesn't seem a good idea.

It'll look better if you use double translation as suggested in Sugarizer [#293](https://github.com/llaske/sugarizer/pull/293).

YJDave:
@ashish0910 The way translation are fixed in #293 is to change it to "language name without any translation + (language name translated to local language)".

ashish0910:
If local language in English:

Screenshot 2019-03-26 at 8 19 06 AM

If local language is French:

Screenshot 2019-03-26 at 8 31 12 AM

If local language is Hindi:

Screenshot 2019-03-26 at 8 19 25 AM

In `()` it shows language name translated to local language and without `()` it shows original language name without being translated to any other language.

ashish0910:
@tarunsinghal92 resolved conflicts

llaske:
Nice but it require to increase size of language selection field because the value is not fully visible now.

Capture d’écran de 2019-04-10 17-16-01

Thanks to resolve new conflicts too.

ashish0910:
@llaske done the fixes

Screenshot 2019-04-10 at 9 26 45 PM

ashish0910:
@llaske can you please review this pr

llaske:
Nice. Thanks.

PR #149, opened 2019-03-26
UX improvement. Fixed links and cursor

NikhilM98:
- In the activities view, since reordering by clicking anywhere on the row to drag feature is replaced with dragging the row by clicking on the vertical ellipsis, we need to make the cursor pointer when the user hovers on the icon such that he can know that there's an action associated with that icon.

- On the navbar, the name of the pages is a clickable link.
They're replaced with divs in this PR.

- The XO icons in some pages have cursor pointer associated with them. It makes the user think that they're buttons instead of images.
 
Sugarizer Dashboard (22)

Fixed:
Sugarizer Dashboard (21)

### DIFF
--- a/dashboard/public/css/main.css
+++ b/dashboard/public/css/main.css
@@ -196,7 +196,6 @@
 
 .color {
   margin: 0 0 5px 0;
-  cursor: pointer;
 }
 
 .color .xo-icon {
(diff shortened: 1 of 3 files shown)


llaske:
Make sense.

PR #153, opened 2019-03-28
[UX Improvement] Logout button

SomtochiAma:
The logout button is a a drop down leading to another log out button which is a bit redundant
Screenshot (36)

The pull request removes one of the buttons


SomtochiAma:
I noticed that, I am working on an implementation that will not affect the desktop view

NikhilM98:
@SomtochiAma I think the UI on the desktop view can be further improved by making the `logout` button more visible to the user. Currently it seems quite difficult to figure out how to logout on the desktop view.

Also, since it's a closed pull request, I'd suggest you to create an issue and assign it to yourself so other contributors would know that you're working on it.

PR #156, opened 2019-03-29
Improve responsive view

NikhilM98:
This PR contains:
- The sidebar is not scrollable on mobile devices.
-> Enabled scroll on mobile view.

- Tables on the dashboard home page were not responsive. The text was going out of the table.
-> Made the table responsive.

Current:
Sugarizer Dashboard (27)

Fix:
Sugarizer Dashboard (28)

### DIFF
--- a/dashboard/public/css/main.css
+++ b/dashboard/public/css/main.css
@@ -194,6 +194,14 @@
   cursor: pointer;
 }
 
+.dashboard-table-title {
+  position: fixed;
+}
+
+.dashboard-table {
+  margin-top: 20px;
+}
+
 .color {
   margin: 0 0 5px 0;
 }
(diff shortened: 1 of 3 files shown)


llaske:
Very cool. Congrats.
BTW I think that the burger menu should be better on the left instead that on the right.
@tarunsinghal92 and @NikhilM98 what do you think?

NikhilM98:
@llaske 
> BTW I think that the burger menu should be better on the left instead that on the right.

When switching from desktop view to mobile view, it feels strange that the sidebar in the mobile view has shifted to the right side. However, when I actually open the dashboard in my mobile phone it feels more comfortable to navigate (With one hand -- Right) since I can easily open the sidebar with my right thumb. If the sidebar is shifted to the left, then I've to stretch my thumb to open the sidebar :sweat_smile:

PR #162, opened 2019-04-02
Removed redundant translation on language change

NikhilM98:
When the language is changed, the page is being translated redundantly on the front end:
1. When the language is changed, the front-end is triggered to translate the language.
2. When the page reloads, it is again translated.

Current behavior:
final_5ca38b708aa7ac001376e2bb_865485

Patch:
final_5ca38b378473240014ed3480_56505

Removed the `l10n` call to update the language on drop-down option change.


llaske:
Good.

PR #165, opened 2019-04-03
Translations for user language

ashish0910:
Translation for Language of users was added in #145 But looks like it is not working now 

Current : 
Screenshot 2019-04-03 at 10 32 58 PM
Screenshot 2019-04-03 at 10 33 07 PM

Expected : 
Screenshot 2019-04-03 at 10 30 50 PM
Screenshot 2019-04-03 at 10 30 44 PM

### DIFF
--- a/dashboard/views/users.ejs
+++ b/dashboard/views/users.ejs
@@ -110,16 +110,16 @@
                                                                                           </script>
                                                                                       <% } %>
                                               <td><%= data.users[i].name %></td>
-                                              <% if(data.users[i].language =='en') {%> <td>English</td> <%} %>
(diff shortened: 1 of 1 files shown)


ashish0910:
> Changes looks good to me.
> 
> > Translation for Language of users was added in #145 But looks like it is not working now
> 
> Looks like the merge commit [72dac76](https://github.com/llaske/sugarizer-server/pull/145/commits/72dac76a9fe0b03843927c2893d5813a5967397f) caused this issue.

@NikhilM98 yes that probably happened because of merge conflicts

llaske:
Ok thanks.

PR #189, opened 2019-07-09
Import/Export users as CSV

NikhilM98:


tarunsinghal92:
@NikhilM98 same for this one too!

NikhilM98:
Screenshots:
- Import
Input
Screenshot from 2019-07-18 11-36-03
Output
Screenshot from 2019-07-18 12-20-50

- Export
Screenshot from 2019-07-18 12-21-55

tarunsinghal92:
Good Job!

llaske:
I think it will be better if export format is the same that the import format.
Plus I think that the id column has no sense.

NikhilM98:
The format is almost the same except for the `_id` column in the output. That column is present in both Import and Export output is it may be useful for confirming the successful creation of the user. The same export file can still be used for imports without any modifications.

PR #192, opened 2019-07-20
Sorting by name and timestamp 

NikhilM98:


llaske:
Very nice.
Is it possible to sort journal by title too?

NikhilM98:
@llaske I've created a PR to allow journal sort by title

PR #195, opened 2019-07-25
Sugarizer Server tutorial

NikhilM98:


dashboard/public/js/tutorial.js
```diff
@@ -0,0 +1,368 @@
...
+			steps: [],
+			keyboard: true,
+			onEnd: function() {
+				tutorial.elements = [];
+			}
+		});
+		if (currentView == "home") {
+			tour.addSteps([
+				{
+					element: "",
```
tarunsinghal92: Quick feedback:
- put steps for each module in a separate file. e.g. `steps/home.json`
- how are you handling multiple languages.

NikhilM98: I'll be using webl10n for translations. I was following Sugarizer tutorial format (https://github.com/llaske/sugarizer/blob/dev/lib/tutorial.js), that why I put it all in a single file.

llaske:
It will be better to launch the tutorial only when the user ask it: by clicking on a help button.

PR #215, opened 2020-01-15
Classroom user search

NikhilM98:
_AddEditClassroom_ View was slowing down for large number of users in the database.
Removed [multiselect.js](http://loudev.com) and replaced it with [jQuery Select2](https://select2.github.io/select2).
Select2 makes AJAX queries for searching on the backend instead of searching on the client-side.

@llaske can you please test this PR before merging.


llaske:
It's very good. The UI is very usable. Thanks.

PR #216, opened 2020-01-16
Classroom inconsistency fix

NikhilM98:
Fixes:
- **Restrict classroom with existing name:** Currently we can create classrooms with the existing name -- This feature may cause issues with seed-users script or upload users as CSV feature.
- **Restrict classroom with empty name:** In Edit Classroom view, we can remove the name of the classroom and save the classroom with empty name.


llaske:
Good job. Thanks.

PR #223, opened 2020-01-30
Private journal checkbox

NikhilM98:
Updated Journal View as requested. @llaske have a look.


llaske:
Good. That's the idea.
BTW are you sure you need to add a new route to get shared entries? why not using the current route like mentioned [here](https://github.com/llaske/sugarizer-server/blob/dev/api/controller/journal.js#L124)?

NikhilM98:
@llaske yes you're right. That route is not required.

llaske:
Hmmm. In fact, the checkbox should not be labeled "Private" but "Shared" instead. Because when the checkbox is checked, only the shared content is displayed, not the private content.

image

PR #224, opened 2020-01-31
Allow teacher login on Sugarizer

### LINKED ISSUE #222: Allow teachers to connect to Sugarizer client
Since https://github.com/llaske/sugarizer/commit/9094711b32ab3dd151954f8d2113be801e139683  Sugarizer allow teachers to login to the app.

BTW it doesn't work because when a teacher is created it doesn't have a private journal and a shared journal. See [here](https://github.com/llaske/sugarizer-server/blob/dev/api/controller/users.js#L493). So the login from Sugarizer app works but it's not possible for a teacher to store something in the Journal or to access to the remote/shared journal.

Peek 26-01-2020 22-11

The idea is to update Sugarizer Server to allow a teacher to login to Sugarizer app with a real Journal.
It means change the teacher creation but also the teacher update and delete.
It's important too to add a new test unit handle it.

NikhilM98:
This patch allows the teacher to log in to Sugarizer.

- [x] The teacher to login to Sugarizer and access their journal entries.
- [x] In Sugarizer-Server, add some way to view teacher journal entries. -- Now teacher entries can be viewed from the journal view.

Fixes #222

### DIFF
--- a/api/controller/auth.js
+++ b/api/controller/auth.js
@@ -82,11 +82,24 @@ exports.login = function(req, res) {
 				}
 			});
 		});
-	} else {
+	} else if (user.role) {
 		var role = user.role || 'student';
 		query.role = {
 			$regex: new RegExp("^" + role + "$", "i")
 		};
+	} else {
+		query['$or'] = [
+			{
+				role: {
+					$regex: new RegExp("^student$", "i")
+				}
+			},
+			{
+				role: {
+					$regex: new RegExp("^teacher$", "i")
+				}
+			}
+		];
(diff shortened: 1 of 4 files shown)


llaske:
Very cool.
BTW I guess we could also add in the Users view the action to see the teacher Journal:

For teachers:
image

For students:
image

NikhilM98:
> I guess we could also add in the Users view the action to see the teacher Journal.

@llaske done.

llaske:
Good job. Thanks.

PR #225, opened 2020-02-09
Journal download

### LINKED ISSUE #220: Add a download icon in Dashboard Journal view
The idea is to add a download action icon on each line of the Journal view.

image

This icon will allow to download the item and save it locally. This feature is the same than the "Copy to device" feature of the Sugarizer Journal.
Depending of the item mime type, the content will be saved as a JSON, JPG, ODT, PDF, ... file.
See the list of supported datatype [here](https://github.com/llaske/sugarizer/blob/dev/lib/util.js#L436).

NikhilM98:
Fixes #220


llaske:
Good work.

Two remarks:

- I suggest to reverse the icon download and the launch icon because it's more important to launch than to download.

image

- It don't work with PDF/ODT/DOCX content (coming from Write activity). I've got a message "No linked activity found". There is a same problem when launching but downloading should be more permissive.

image

llaske:
That's fine.

But there is an issue with encoding of TXT file:

image

Guess it's considered as binary.

NikhilM98:
@llaske it's working fine on my side. How did you create this text file?
I uploaded a .txt file by using the _Import from device_ feature of the Sugarizer app.

Screenshot from 2020-02-10 23-22-45

llaske:
Hmmm. You're right, the TXT handling was different when imported into Journal and when generated by Write activity. I've fixed the bug [here](https://github.com/llaske/sugarizer/commit/bee87e1fcd5968758be226cf38636c98733fa298). BTW you will have to update your code to handle the plain text format. Just add the line:

`text = content;`

Just after [this line](https://github.com/llaske/sugarizer-server/pull/225/files#diff-48ada1f0866d4f36fd278ee459dfb70eR926). And it should work.

There is another issue: the Tooltip is not translated on the Download icon. Here it should be "Télécharger" like mentioned in the .INI file.

Capture du 2020-02-10 22-29-49

NikhilM98:
@llaske done!

llaske:
Good job. Thanks.

PR #229, opened 2020-02-15
Journal Upload

### LINKED ISSUE #221: Add an upload button in Dashboard Journal view
The idea is to add a upload button in the Journal view to provide a way for an admin or a teacher to add content in a private Journal or in the shared Journal.

image

This feature is the same than the "Copy from device" feature of the Sugarizer Journal.
Depending of the file extension the content will be saved in the Journal a document or as an activity instance. See [here](https://github.com/llaske/sugarizer/blob/dev/js/journal.js#L1148) and [here](https://github.com/llaske/sugarizer/blob/dev/lib/util.js#L548) for more.

NikhilM98:
Fixes #221

### DIFF
--- a/dashboard/controller/activities/launch.js
+++ b/dashboard/controller/activities/launch.js
@@ -71,8 +71,8 @@ module.exports = function launch(req, res) {
 				lsObj['sugar_settings'].name = user ? user.name : req.session.user.user.name;
 				lsObj['sugar_settings'].color = 128;
 				lsObj['sugar_settings'].colorvalue = (user && user.color) ? user.color : req.session.user.user.color ? req.session.user.user.color : {
-					"stroke": "#808080",
-					"fill": "#ffffff"
+					"stroke": "#005FE4",
+					"fill": "#FF2B34"
 				};
 				lsObj['sugar_settings'].connected = false;
 				lsObj['sugar_settings'].language = user ? user.language : req.query.lang ? req.query.lang : "en";
--- a/dashboard/controller/activities/launch.js
+++ b/dashboard/controller/activities/launch.js
@@ -68,12 +68,15 @@ module.exports = function launch(req, res) {
 			getUser(req, body.entries[0].metadata.user_id, function(user) {
 
 				//set fields
-				lsObj['sugar_settings'].name = user.name;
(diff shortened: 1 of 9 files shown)


llaske:
Good work. It seems to work.
BTW I'm not a big fan of the button.
May be we could just use an icon. For example the reverse of the download icon.

image

NikhilM98:
@llaske what do you think about this button?
Screenshot from 2020-02-16 21-37-30

llaske:
It's very good. But look like you've added it only in Shared journal:

image

Plus, when I'm upload a content in the shared journal, I can't reopen it from the dashboard. Look like the buddy colors are missing.

NikhilM98:
@llaske actually there was a check for empty colors:
https://github.com/llaske/sugarizer-server/blob/f0852590c48fa5f7282ff5c0be3498312793e507/dashboard/controller/activities/launch.js#L73-L76

If the color is present in the Journal, it takes the color defined in the journal.
Else if the user is a teacher, opens the journal with the colors of that teacher.
Else it opens the journal with the default colors.

I tested the feature with the Shared Notes activity and it seems to work fine. I'll try to further test this feature with other activities.

NikhilM98:
@llaske fixed!

dashboard/controller/activities/launch.js
```diff
@@ -70,7 +70,7 @@ module.exports = function launch(req, res) {
 				//set fields
 				lsObj['sugar_settings'].name = user ? user.name : req.session.user.user.name;
 				lsObj['sugar_settings'].color = 128;
-				lsObj['sugar_settings'].colorvalue = user ? user.color : req.session.user.user.color ? req.session.user.user.color : {
+				lsObj['sugar_settings'].colorvalue = (user && user.color) ? user.color : req.session.user.user.color ? req.session.user.user.color : {
 					"stroke": "#808080",
```
llaske: I suggest to use the standard blue/red `{stroke:"#005FE4",fill:"#FF2B34"}` colors instead of gray.

NikhilM98: Updated default colors

llaske:
Good job. Thanks.

PR #230, opened 2020-02-20
Updated Locale

NikhilM98:


llaske:
I'm not a big fan of automatic translation. 
I guess you're a native Hindi speaker so that's fine for your changes in Hindi but it's better to let Spanish text in English waiting for a translation by a native speaker.

NikhilM98:
I think having automatic Spanish translations is better than not having translations at all. Most of the previous Spanish texts in Sugarizer-Server were translated using translation tools. Maybe we can create an issue in the Sugarizer-Server for a request to improve existing automatic Spanish translations.

If you insist, I can remove the automatic Spanish translations from this PR. What do you think?

llaske:
Hmmm. My strategy is to let native speaker localize contents in its language. So my preference is to let all strings in English because the current translation platform (https://translate.sugarizer.org) give a warning when the localized string is the same that the English one. So it's easy for translators to know what should be localized.

So yes, I suggest to remove Spanish localization from this PR. May be we could even fully remove Spanish localization from the dashboard.

llaske:
Nice. Thanks.

PR #234, opened 2020-03-03
Added and fixed some Hindi translations

dhruvmisra:
Added Hindi strings for Import and Export Users and fixed one more.

Before:
localeIssue

After:
localeFix


llaske:
@tarunsinghal92 or @NikhilM98 could you review?

dashboard/public/l10n/locales.ini
```diff
@@ -776,7 +776,7 @@ ClassroomUpdated=कक्षा {{name}} को सफलतापूर्व
 ClassroomCreated=कक्षा {{name}} सफलतापूर्वक बनाया गया है!
 ClassroomDeleted=कक्षा {{name}} को सफलतापूर्वक हटा दिया गया है!
 DoYouWantDeleteClassroom=क्या आप कक्षा {{classroom}} को हटाना चाहते हैं?
-ShortForBytes=ब
```
NikhilM98: This should not be changed. The data is in Bytes not Bits.

dashboard/public/l10n/locales.ini
```diff
@@ -817,8 +817,8 @@ HowManyUsersActiveLastWeek=पिछले सप्ताह में कि
 HowManyUsersActiveLastMonth=पिछले महीने में कितने उपयोगकर्ता सक्रिय हैं
 HowManyUsersActiveLastYear=पिछले वर्ष में कितने उपयोगकर्ता सक्रिय हैं
 CountStudents=छात्र गणना
-ImportUsers=Import users
-ExportUsers=Export users
+ImportUsers=उपयोगकर्ता आयात करें
```
NikhilM98:
```suggestion
ImportUsers=आयात उपयोगकर्ता
```
Shorter button text

dashboard/public/l10n/locales.ini
```diff
@@ -817,8 +817,8 @@ HowManyUsersActiveLastWeek=पिछले सप्ताह में कि
...
+ExportUsers=उपयोगकर्ता निर्यात करें
```
NikhilM98:
```suggestion
ExportUsers=निर्यात उपयोगकर्ता
```

dashboard/public/l10n/locales.ini
dhruvmisra: Thanks for the review. Although shouldn't this follow the common trend of using बी for Bytes? For example in ShortForKilobytes=केबी and ShortForMegabytes=एमबी.

llaske:
@NikhilM98 could you check the fix?

dashboard/public/l10n/locales.ini
NikhilM98: If you use बी for Byte then it could cause confusion between Bit and Byte.
@ashish0910 what do you think?

ashish0910: I think ब makes more sense .

dhruvmisra: Fixed in latest commit.

llaske:
Nice. Thanks.

PR #241, opened 2020-03-10
[Issue #240 fix] Fixed selected text localization

dhruvmisra:
Fixes #240 

selectedLocalizeFix

PR #242, opened 2020-03-11
[Issue #142 fix] Fixed query retention after language switch

### LINKED ISSUE #142: Page navigation problem on language switching
After switching the language the reload takes place and forgets the query of the user and redirects to default situation which is not good .

Example :

ezgif com-video-to-gif 

Similar behaviour takes place in case of journal also .

dhruvmisra:
Fixes #142.
On localization, I'm taking the current query and modifying it to update the language. I handled all three cases:
1. Query string empty (appending the lang)
2. Query non-empty but does not contain lang parameter (adding lang)
3. Query non-empty and contains lang parameter (Updating lang using regex)

langSwitchNavFix2

@NikhilM98 I believe this solution would not have any unexpected results as it's basically similar to refreshing the page, which was already taking place in case of language switching. The other query parameters stay the same regardless of the language chosen, hence not impacting the query search.

Please review @llaske, @NikhilM98 and @ashish0910.

### DIFF
--- a/dashboard/public/js/main.js
+++ b/dashboard/public/js/main.js
@@ -639,8 +639,7 @@ function onLocalized() {
 			if(searchQuery.length == 0) {
 				//query empty
 				searchQuery = '?lang=' + lang.value;
-			}
-			if(searchQuery.indexOf('lang=') != -1) {
+			} else if(searchQuery.indexOf('lang=') != -1) {
 				// query contains the 'lang=' parameter
 				searchQuery = searchQuery.replace(/lang=[a-z][a-z]/, 'lang=' + lang.value);
 			} else {
--- a/dashboard/public/js/main.js
+++ b/dashboard/public/js/main.js
@@ -635,7 +635,19 @@ function onLocalized() {
 		}
 		lang.onchange = function() {
 			localStorage.setItem("languageSelection", this.value);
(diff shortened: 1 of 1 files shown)


dashboard/public/js/main.js
```diff
@@ -635,7 +635,19 @@ function onLocalized() {
 		}
 		lang.onchange = function() {
 			localStorage.setItem("languageSelection", this.value);
-			location.href = window.location.pathname + "?lang="+lang.value;
+			var searchQuery = location.search;
+			if(searchQuery.length == 0) {
+				//query empty
+				searchQuery = '?lang=' + lang.value;
+			}
+			if(searchQuery.indexOf('lang=') != -1) {
```
NikhilM98: It should be else if.

dhruvmisra: Right, made the change.

llaske:
It don't solve the issue, there is still a mix between english and native language in the combo.

dhruvmisra:
@llaske I'm afraid this PR is for a different issue. The issue you mentioning has been solved in #241.

llaske:
@dhruvmisra oops, you're right.
Nice job on this PR and on the https://github.com/llaske/sugarizer-server/pull/241

PR #244, opened 2020-03-25
Added title to charts in statistics view

NikhilM98:
Charts in the Statistics View were missing Title.

Before:
Screenshot from 2020-03-26 00-15-28

After:
Screenshot from 2020-03-26 00-03-48


llaske:
I don't see the interest to add both the Title and the graphic Name.
Plus Title field has limitation (doesn't support punctuation, doesn't support accentuated characters, ...) today, so it's not very practical.
In my opinion we should have only one field Title with a default value equal to the current graphic Name.

NikhilM98:
@llaske if we have only title field with default value equal to the current graphic name, then it will cause problem with translation. The graphic field supports translation but the title does not. If we add a title with the default value as the graphic name, then it won't be translated on language change.
What do you think?

llaske:
@NikhilM98 good remark. But it's an issue only for graphics created by default. For graphics created by users, we could expect than the user use its own language.
May be we could consider than if the user leave the title blank the default name will be use (and will be localized)?

NikhilM98:
> we could expect than the user use its own language

Yeah. I was thinking the same. It's expected that the user will use his own language.

> May be we could consider than if the user leave the title blank the default name will be use (and will be localized)?

Sounds good

NikhilM98:
@llaske I have made the changes. Please have a look.

llaske:
That's nice like that. Thanks.

PR #246, opened 2020-03-26
Added tutorial for User Edit and Profile view

dhruvmisra:
I noticed the tutorial buttons on Profile and User Edit views didn't open anything. Added a tutorial for them. Please review and add French translations @llaske.


NikhilM98:
I have tested the PR. Here are my remarks:
- Initially, tutorials were not added to the AddEditUser, AddEditClassroom, and AddEditChart views because those views are simple self-explanatory forms. We don't need tutorials to explain deployment administrators how to fill forms.
- There is an issue. After completing half of the tutorial on the Add User page. If we try to open the Edit Users page, the tutorial starts from Step-8.
- If you're adding tutorials for the Forms, I think it'd be better to add the tutorial for AddEditClassroom and AddEditChart views too in this PR.
- Can you use "colors" instead of "colours" in the locale to maintain consistency?

dhruvmisra:
Hi @NikhilM98, thanks for reviewing.
* I believe the tutorials explain what the field stores and explains the unique sections such as Classroom selection. It doesn't teach administrators how to fill a form, rather it makes it clear what the fields signify to the system.
* I'm not able to replicate this issue, can you please tell me the steps to do the same?
* Sure thing, I'll add a tutorial for those views as well.
* Sure, I'll make the change right away.

NikhilM98:
1. Clear the application cache.
2. Open Add Users View finish the tutorial for that.
3. Open Edit Users View. The tutorial jumps to Step-8 (Last Seen)

dhruvmisra:
@NikhilM98 Fixed the issue.

NikhilM98:
Just a small difference from the usual behaviour. Normally when the Sugarizer-Server freshly starts (Empty Application Cache), the tutorials are shown automatically for the first time. If you refresh the browser window in between of the tutorial, the tutorial continues from the last step. However, on addEditUsers view, it starts from the beginning on refreshing.

dhruvmisra:
@NikhilM98 As the two views reuse the same EJS file, I have added a *view* property to localStorage to identify if we are still on the same view. Through this, I can continue the tutorial from the last step if the page was refreshed and restart it if it's a new view.
I've also added tutorials for *addEditClassroom* and *addEditChart*. Please review.

NikhilM98:
Some translations in the locale.ini are missing (after line 661).

dhruvmisra:
Fixed in the latest commit.

llaske:
Nice. Thanks.

PR #249, opened 2020-03-29
Added multi-delete and multi-download features

### LINKED ISSUE #245: Add multiple actions to Journal and Users view
The entries in the Journal view for a user shows a list with individually dispatch-able actions (Launch/Download/Delete). Wouldn't it be a good feature to have check boxes similar to Sugarizer journal to perform batch operations (mainly download and delete). It would speed up the process of download and deletion. What do you say @llaske?

NikhilM98:
Fixes #245

Screenshot from 2020-03-29 16-54-04


NikhilM98:
Update: Added multi-delete feature in users view.

llaske:
That's good.
BTW I'm not a big fan of the popup when no item is selected.

image

It's the only time in the dashboard where there is a such popup.

I suggest several options:
- Disable the button when no item is selected
- Hide the button when no item is selected
- Display a "server popup"

What do you think?

NikhilM98:
@llaske
I think hiding showing the button may cause little confusion (with tutorial and UX).
Disabling the button is a good idea but the user may not understand why that button is disabled.
I have displayed server popup if no user was selected. That way the user can understand what went wrong.

llaske:
That's nice. Thanks.

PR #251, opened 2020-04-03
Documentation fixes for deploying to GCP

ksraj123:
Documentation fixes
Paste was spelled wrong and the commands to be pasted were missing.
Some commands and steps are different for GCP so they need to be defined here for easy deployment.


ksraj123:
@NikhilM98 Yeah I tried that but I was getting permission denied even with sudo.
Screenshot from 2020-04-04 05-54-14

ksraj123:
@NikhilM98  However specifying the output file with `-o` flag does the trick.

PR #252, opened 2020-04-04
Removed select2 data preload

NikhilM98:
The preload-data line is causing an error in the console and is not required.

Screenshot from 2020-04-04 23-35-25


llaske:
You're right. Thanks.

PR #255, opened 2020-04-07
Added All Classrooms option in Users View

NikhilM98:
- Added All Classrooms option in the Users View. The Select Classroom had no option to reset the selected option.
Screenshot from 2020-04-08 00-30-13

- Reduced code redundancy.
- Reduced window event listeners (on localized).


llaske:
Oh you're right.
Thanks.

Issue #256, opened 2020-04-09
Simplifying deployment to GCP using Ansible


ksraj123:
I think deployment of sugarizer-server onto the Google Cloud Platform makes sense for many schools or other organisations trying to setup sugarizer deployment themselves as they provide free credit initially.

But the process for deployment is lengthy and may seem complex or intimidating to non-developers.

The UI of Google Cloud Platform is not very user friendly for non-developers in my opinion but the current deployment process requires to user to create and configure vm instances and firewalls themselves by navigating though the UI and then ssh into the vm to execute a script.

I think deployment will be much simpler if an ansible package is realised which would create a vm instance, firewall and execute required scripts with a single command.
The user will provide essential information like project id and authentication details and will be able to modify other variables like machine type, os etc if needed.

I am working on this, will open a PR soon.

Please share suggestions.

llaske:
Hmmm. Ansible is a good idea. It's on our TODO list for GSoC 2020 project [Sugarizer School Portal](https://github.com/sugarlabs/GSoC/blob/master/Ideas-2020.md#sugarizer-school-portal).
But may be we could imagine first a generic Ansible package to deploy Sugarizer Server, then a specific package to deploy it on GCP.

NikhilM98:
I think Ansible package in the Sugarizer School Portal project is to bring the K8s Infrastructure up. It is not for installing Sugarizer Server using Ansible. Even in the Sugarizer School Portal a Helm chart will be created to install Sugarizer-Server. Currently, Sugarizer-Server can be very easily installed using Docker so I think there is no need for Ansible. One drawback is that we have to depend on ansible for simple setup.

ksraj123:
@llaske I agree with @NikhilM98 , for simple docker deployment on localhost an ansible package will be as good as a simple script. 

However, I think at least for deployment to GCP specifically, an ansible package would make sense for reasons mentioned before.

#257 - proposes an ansible package for deployment to GCP 
#259 - package for simple deployment to localhost

I am also working upon the [sugarizer school portal ansible package](https://github.com/ksraj123/sugarizer-server-k8s-ansible) which creates a cluster onto Google Kubernetes Engine and installs the helm chart for sugarizer-server. Its a work in progress, would love to have your feedback on the current build.

llaske:
@ksraj123 @NikhilM98 the idea is not to replace the current way to deploy Sugarizer Server with docker.

But ansible could be useful in two ways:
[...]

PR #257, opened 2020-04-10
Ansible package for easy deployment on GCP

ksraj123:
Fixes Issue #256 

This ansible package simplifies the current deployment method for deploying sugarizer-server on vm instances on google compute engine.
It creates and configures a vm instance and firewall, installs dependencies and deploys sugarizer-server onto the vm instance, it also prints the external IP of the vm instance where the deployment can be accessed.
 
This could be tested with steps described in `ansible/docs/ansible-deployToGcp.md` of this PR.


ksraj123:
@NikhilM98 Thanks for reviewing. I have made the suggested changes. Deprecated module `gce` has been replaced with `gcp_compute_instance`

Yes, ansible/deploy-to-gcp does not have any direct dependence upon Sugarizer Server Code that is why only this specific directory is downloaded to the user's local machine by steps mentioned in ansible/docs as we do not want the user to unnecessarily download the entire repository.

`get_start_script` uses code from github as this script is being executed in our vm instance on GCP not the user's local machine where the code might already be present.

Ansible packages for other deployments methods too can be created in future and they can all be inside the ansible folder, I think it will promote gradual transition to using ansible as the preferred way to setup sugarizer-server specially by non-developers. However separating it into another repository is also a good option.

Requesting suggestions @llaske

NikhilM98:
> `get_start_script` uses code from github as this script is being executed in our vm instance on GCP not the user's local machine where the code might already be present.

Have you tested providing `startup-script` directly instead of `startup-script-url`?
Reference: [link](https://cloud.google.com/compute/docs/startupscript)

ksraj123:
@NikhilM98 Thanks for suggestion. `startup-script` was not used initially as providing path to script file as described in the reference link did not work, maybe because I could not seem to find `gcp_compute_instance` counterpart of `--metadata-from-file` flag of `gcloud`.  We could have installed google cloud sdk on the local machine of the user and used the gcloud command as in the reference link but its relatively heavy weight or provide the content directly but it would clutter our playbook. However another work around is used in which the script is defined in a yaml file and variables from that file is included in our playbook. This approach also ensures that user does not execute the startup script which is meant to be executed on the vm instance on his local machine.

NikhilM98:
It looks better now. The reference link was added to point out that we can use `startup-script` as a metadata key instead of `startup-script-url` to provide the startup instructions.

Btw the file `4.png` is uploaded but not used.

ksraj123:
@NikhilM98 the unused image `4.png` has been removed.

llaske:
@mikklfr could you review this and see if it could be helpful for Sugarizer School Portal project.

mikklfr:
Hi!
This is nice, this Ansible recipe will allow to easily create and fill up an instance.

That's not linked with the school project technically as this approach is not about sharing a pool of resources for many schools but about setting up a single deploy.

But that is nice and could simplify deployments for some.

llaske:
Okay.

@ksraj123 could you:
- Reduce images size, currently 1920x880, should be near 800x400
- Could you put all documents into `docs` directory instead of the `ansible/docs` directory
- Could you reference - or even include - your documentation into existing GCP documentation. May be you could describe it as an alternative to manual install
- Use relative path `llaske/sugarizer-server/dev` instead of `ksraj123/sugarizer-server/patch-1`

ksraj123:
@llaske Suggested changes made, please review.

llaske:
Nice. Thanks.

PR #258, opened 2020-04-10
Added multi-delete in classroom view

NikhilM98:
Added multi-delete feature in the classroom view (following #249)


llaske:
Nice. Thanks.

PR #259, opened 2020-04-12
ansible package for deployment to localhost

ksraj123:
After discussions in issue #256 a simple ansible package for generic deployment to localhost is created, #257 proposes a package for deployment to GCP.
Docker method is adopted here for deployment, please following steps documented in `ansible/docs/ansible-deploy-to-localhost.md` to test this. 
But the package is as good as executing a script.
Suggestions requested on what more can be done.


llaske:
Hmmm. I don't see the interest to use Ansible just to launch Docker. Was not a good idea.

PR #260, opened 2020-04-12
Replace request with superagent

### LINKED ISSUE #254: "request" library has been deprecated
The _[request](https://github.com/request/request)_ library has been deprecated. According to [Issue #3142](https://github.com/request/request/issues/3142) in its official repository, _request_ will stop accepting new features and it will stop considering breaking changes.
https://github.com/llaske/sugarizer-server/blob/6319475cb73cc2135d9cf3bcc08e677fd422e8b8/package.json#L27

This library is used in Sugarizer Server dashboard controllers to make HTTP requests.

We might need to replace _request_ with [alternatives](https://github.com/request/request/issues/3143).

I think [superagent](https://www.npmjs.com/package/superagent) would be a good alternative as it supports callbacks.

NikhilM98:
Fixes #254

Replacing `Request` with `Superagent`.

Commit [14dd7](https://github.com/llaske/sugarizer-server/commit/14dd7cb01e282ea2baee1842b7cea8e9c68381e7) shows the modification we need to do to replace `request` with `superagent` in each file.

The changes are minimal.
- Both libraries use a callback function.
- HTTP Request Method, Headers and Query String can be set easily and the inside logic remains the same.
- Just have to replace `body` with `request.body` inside the function.

@llaske Should we go ahead with `Superagent` and remove `request` dependence completely?

### DIFF
--- a/dashboard/controller/activities/index.js
+++ b/dashboard/controller/activities/index.js
@@ -1,5 +1,5 @@
 // include libraries
-var request = require('request'),
+var superagent = require('superagent'),
 	common = require('../../helper/common'),
 	fakeLaunch = require('./fakeLaunch'),
 	launch = require('./launch');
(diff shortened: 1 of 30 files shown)


NikhilM98:
Finished replacing `request` with `superagent`. I have manually tested all the components.
@llaske, please have a look.

llaske:
Very good. Thanks.

PR #261, opened 2020-04-14
Modifications to Mongo Url for kubernetes compatibility

ksraj123:
In kubernetes deployment our sugarizer-server and mongodb containers will be in different pods.
Our service for mongodb pods will be assigned a DNS record for the following form
`<serviceName>.<namespace>.svc.cluster.local` which will be used to connect to the mongo pods.
I understand that this can be done by providing the DNS record in `database.server` of a new `kubernetes.ini` file having most values for most the properties same as `docker.ini` but the issue is that `serviceName` and `namespace` are not static strings and their values would have to be passed as environment variables in our pod manifest file. So passing a single environment variable `MONGO_URL` seems to be the least invasive and more simple approach.

A sample deployment on Google Kubernetes Engine incorporating changes in this PR can be accessed on - http://34.84.103.197:8080/

Details on how the env var will be passed in the deployment manifest file can found [here](https://github.com/ksraj123/sugarizer-server-k8s-ansible/blob/24a2630f7c61dfdcf8d17b4663e9cd4234cd063e/sugarizer-chart/templates/server_deployment.yaml#L40)

### DIFF
--- a/config/wait4db.js
+++ b/config/wait4db.js
@@ -40,6 +40,6 @@ exports.waitConnection = function(settings, callback) {
 
 function createConnection(settings) {
 	return new mongo.MongoClient(
-		'mongodb://'+settings.database.server+':'+settings.database.port+'/'+settings.database.name,
+		(process.env.MONGO_URL || 'mongodb://'+settings.database.server+':'+settings.database.port)+'/'+settings.database.name,
 		{auto_reconnect: false, w:1, useNewUrlParser: true, useUnifiedTopology: true });
 }


NikhilM98:
I appreciate your efforts, but I think it's too early to create this PR. Reasons:
- Sugarizer-Server dev branch is quite stable right now. I don't think that it would be a good idea to continuously merge changes regarding Sugarizer School Portal with the latest dev branch. It might take more than 4 months from now for the first release of Sugarizer School Portal.
- I think there is a be a better way of connecting with MongoDB in Kubernetes without using the environment variable. So this change is not needed. (This can be done by modifying the `sugarizer.ini` in the chart itself -- change the MongoDB server) 
- The exact workflow and implementation will be finalized during the community bonding period and weekly meetings. Every student has their own implementation in mind, the student for the project has not been selected yet. If we merge these changes right now, then we might have to revert them later.

A better way to proceed right now will be to maintain the changes in your fork and discuss the impact and changes with the community. I think it will be better to continue improving/fixing Sugarizer/Sugarizer-Server for the time being.

ksraj123:
@NikhilM98 I agree with you on keeping changes regarding surgairzer school portal to my fork. I try to open PRs only for changes which are not strictly related to sugarizer school protal and I try to make then as subtle as possible. I thought the environment variable proposed in this PR could be useful in other situations as well. I am not sure how changing `sugarizer.ini` in the chart would work for my implementation but surely we could figure out something better than this PR. I have another open PR #262 would love to have your opinion on the same.
Thank you.

mikklfr:
Hi everyone, using the env is generally preferred over conf files but I suggest we wait for the student to be chosen.
Having two systems for the same result (passing mongodb infos) could lead to bugs.
Kubernetes allows to mount file in the pod so it's possible to avoid the need for env.
Maybe that's the way we should go, or not, but let's wait for the project to begin.
Cheers,
Michael

ksraj123:
Thanks @mikklfr for reviewing.

PR #263, opened 2020-04-14
Handling sessions across multiple instances, kubernetes compatibility

### LINKED ISSUE #262: Handling sessions across multiple instances, kubernetes compatibility
When there are more than one instance of sugarizer-server in a kuberentes deployment then the user may get redirected to login page when he either tries to log in or access any route after login.
Peek 2020-04-14 20-13

A sugarizer deployment on Google Kubernetes Engine with 3 instances of sugarizer-server can be accessed here to explore the issue - http://34.84.103.197:8080/dashboard/login  
A sample helm chart is developed which is being user for deployment, it can be accessed [here](https://github.com/ksraj123/sugarizer-server-k8s-ansible/tree/master/sugarizer-chart)
A default admin account has been added for testing purpose, 
username - defaultuser
password - password

Currently no session store is used so it defaults to `MemoryStore' . In kubernetes environment there are multiple instance of sugarizer-server and the request can be routed to any one these, I guess this is the root cause of the issue as one instance will not be aware of a session in another instance's memory. 

Also it is [not recommend](https://github.com/expressjs/session#sessionoptions) to use default `MemoryStore` as it is prone to memory leaks.

ksraj123:
Fixes #262 

Since mongodb is already being used `connect-mongo` session store is used. This enables sessions to work across multiple instances and does not affect normal functionality.

@NikhilM98 @llaske The changes in this PR are not strictly related to sugarizer school portal as it is [not recommend](https://github.com/expressjs/session#sessionoptions) to use default `MemoryStore` as it is prone to memory leaks. Also, this issue will be apparent in any deployment scenario where there are more than 1 instance of sugarizer-server to whom requests can be routed to.

A deployment on Google Kubernetes Engine, incorporating changes in this PR with 3 instances of sugarizer-server can be accessed here for testing purposes - http://104.198.93.25:8080/dashboard/login
A default admin account has been added for testing purpose, 
username - defaultuser
password - password

fixed session Peek 2020-04-14 21-05

### DIFF
--- a/config/main.js
+++ b/config/main.js
@@ -5,11 +5,12 @@ var express = require('express'),
 	logger = require('morgan'),
 	flash = require('express-flash'),
 	expressValidator = require('express-validator'),
-	cors = require('cors');
+	cors = require('cors'),
+	mongoStore = require('connect-mongo')(expressSession);
 var artificialDelay = 0;
 
 // configure app settings
-module.exports = function(app, ini) {
+module.exports = function(app, ini, db) {
 
 	// add an artificial delay for debug
 	if (artificialDelay) {
@@ -26,7 +27,10 @@ module.exports = function(app, ini) {
 			maxAge: parseInt(ini.security.max_age)
 		},
 		resave: true,
-		saveUninitialized: true
+		saveUninitialized: true,
+		store: new mongoStore({
+			db: db
+		})
 	}));
 
 	//include body parser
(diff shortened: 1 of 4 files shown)


NikhilM98:
Sugarizer Server has a web server running on port 8080 and presence running on port 8039.
These changes will make Sugarizer-Server use MongoDB to store sessions which were previously stored in the memory. However, the presence data will remain to be stored in the memory. So if you increase the number of replicas, the users will continue to have issues with the presence.

ksraj123:
Thanks for reviewing, hope it works out well in the final build for sugarizer school portal

PR #264, opened 2020-04-14
Fix server error flashing

NikhilM98:
`flash-errors` was not included in some views where server popup is used due to which server popup is not displayed.

Steps to reproduce: When you try to Add/Edit/Delete a Chart, server popup will not be displayed.

Popup is displayed after applying the patch:
Screenshot from 2020-04-15 00-55-53


llaske:
You're right. Thanks.

PR #266, opened 2020-04-21
Updated signup function to validate name

dhruvmisra:
Fixes llaske/sugarizer#580.

This PR is analogous to llaske/sugarizer#786. Handles the new `beforeSignup` flag and adds `validateUsername` function similar to `validateUser` to check for the uniqueness of username.

### DIFF
--- a/api/controller/auth.js
+++ b/api/controller/auth.js
@@ -126,7 +126,7 @@ exports.login = function(req, res) {
 /**
  * @api {post} auth/signup/ Signup User
  * @apiName Signup User
- * @apiDescription Add a new user (Admin or Student). Return the user created.
+ * @apiDescription Add a new user (Admin or Student). Return the user created. Additionally, before signing up, check if the new name does not already exist.
  *
  * For security reason, call to signup for an Admin is only allowed from the server address.
  * @apiGroup Auth
@@ -144,6 +144,7 @@ exports.login = function(req, res) {
  * @apiSuccess {String} private_journal Id of the private journal on the server
  * @apiSuccess {String} shared_journal Id of the shared journal on the server (the same for all users)
  * @apiSuccess {Number} timestamp when the user was created on the server
+ * @apiSuccess {Boolean} beforeSignup Flag to check for name validity
  *
  * @apiSuccessExample Success-Response(Student):
  *     HTTP/1.1 200 OK
@@ -161,6 +162,12 @@ exports.login = function(req, res) {
  *       "shared_journal": "536d30874326e55f2a22816f",
  *       "timestamp": 1423341000747,
  *       "_id": "5569f4b019e0b4c9525b3c97"
+ *    }
+ *
+ * @apiSuccessExample Success-Response(Before signup):
+ *     HTTP/1.1 200 OK
+ *     {
+ *       "exists": false
  *    }
  *
  * @apiSuccessExample Success-Response(Admin):
--- a/api/controller/auth.js
+++ b/api/controller/auth.js
@@ -196,7 +196,9 @@ exports.signup = function(req, res) {
 
 function validateUsername(name, callback) {
 	users.getAllUsers({
-		'name': name
+		'name': { 
+			$regex: new RegExp("^" + name + "$", "i") 
+		}
 	}, {}, function(users) {
 		if (users.length > 0) {
 			callback(false);
--- a/api/controller/auth.js
+++ b/api/controller/auth.js
@@ -178,11 +178,14 @@ exports.signup = function(req, res) {
 
 	var user = JSON.parse(req.body.user);
 	if(user.beforeSignup) {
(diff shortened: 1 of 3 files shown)


api/controller/auth.js
```diff
@@ -176,8 +176,34 @@ exports.login = function(req, res) {
...
+		});
+	} else {
+		//insert the user using the same logic but without auth
+		users.addUser(req, res);
+	}
+};
+
+function validateUsername(name, callback) {
+	users.getAllUsers({
+		'name': name
```
NikhilM98, line 199: Use string with regex similar to this instead of directly passing the name.

https://github.com/llaske/sugarizer-server/blob/a5232f33f28341173fcf551529555c342e4ad08b/api/controller/auth.js#L68-L70

api/controller/auth.js
```diff
@@ -176,8 +176,34 @@ exports.login = function(req, res) {
...
+	}, {}, function(users) {
+		if (users.length > 0) {
+			callback(false);
```
NikhilM98, line 202: Return user instead of boolean so that the function can be reused in future.

https://github.com/llaske/sugarizer-server/blob/a5232f33f28341173fcf551529555c342e4ad08b/api/controller/auth.js#L183-L195

api/controller/auth.js
```diff
@@ -176,8 +176,34 @@ exports.login = function(req, res) {
  **/
 exports.signup = function(req, res) {
 
-	//insert the user using the same logic but without auth
-	users.addUser(req, res);
+	var user = JSON.parse(req.body.user);
+	if(user.beforeSignup) {
+		validateUsername(user.name, function(unique) {
+			if(unique) {
+				res.send(true);
```
NikhilM98, line 183: Send a more descriptive response instead of just `true`. Something like `{exists: false}`.

api/controller/auth.js
dhruvmisra: Added this in b7343ef.

api/controller/auth.js
dhruvmisra: Made the change in 1a7cc82.

api/controller/auth.js
dhruvmisra: Made the change in 1a7cc82.

llaske:
Using a flag could be a work around but it seems that a good practice to test existence of a REST resource should be done by a `HEAD` call (instead of `GET`). See for example [here](http://www.pragmaticapi.com/blog/2013/02/14/restful-patterns-for-the-head-verb/).
I think it's a better idea that using a flag.

@dhruvmisra @NikhilM98 what do you think?

NikhilM98:
Yeah, I was also thinking that having another end-point to check for the existence of the user would be a nice idea.
My points are:
- If we create another route in [route.js](https://github.com/llaske/sugarizer-server/blob/dev/api/route.js) file to check for a user's existence, we will be creating an unprotected route whose only purpose will be to check for the availability of the username white creating a user, it won't have any other use.
- `/api/v1/users` API already exists to get user or check user availability in the Sugarizer-Server, so the new route would not be needed by the Sugarizer-Server.

About `HEAD` call instead of `GET`:
I think using `HEAD` call instead of `GET` to check resource availability is a good idea, it will help improve sugarizer-server REST API standard over time.

However, I found out on [Express docs](https://expressjs.com/en/api.html#router.METHOD) that:
> The `router.get()` function is automatically called for the HTTP HEAD method in addition to the GET method if `router.head()` was not called for the path before `router.get()`.

This means that if we create a route with `router.get()` method in Sugarizer-Server, we can use both `HEAD` and `GET` to call that API.

---

Using `HEAD` call to check for the resource (user) availability seems a good idea. But I'm not sure that we should create a new API for that or not.

---

I have also replied on [Sugarizer PR-786](https://github.com/llaske/sugarizer/pull/786) addressing the issue with the backwards compatibility.

llaske:
Hmmm. Thanks @NikhilM98 for detailed explanation.

> I have also replied on Sugarizer PR-786 addressing the issue with the backwards compatibility.

@dhruvmisra, let's try that. If it works, we could avoid to create a new end-point or create a HEAD call.

llaske:
It works well.
@dhruvmisra, could you update API documentation to add the mention of this new parameter?

dhruvmisra:
Made the documentation changes to `api/controller/auth.js`, `docs/www/api_data.js` and `docs/www/api_data.json`. Let me know if there are any changes required.

llaske:
Thanks.

FYI, sound like you've updated HTML documentation by hand. You must not. See [here](https://github.com/llaske/sugarizer-server/tree/dev#server-api) to understand how to generate API documentation. I've done it and fix a small issue on the new parameter declaration (should be optional).

dhruvmisra:
Okay thanks. I'll keep that in mind from next time.

Issue #187, opened 2019-06-27
Error on expressValidator when running Server


llaske:
When the server is run, an error is displayed :

```
lionel@lionel-sugarizer:~/src/olpc/sugarizer-server$ node sugarizer.js
FATAL ERROR, uncaught exception 'expressValidator is not a function'
TypeError: expressValidator is not a function
    at module.exports (/home/lionel/src/olpc/sugarizer-server/config/main.js:46:10)
    at /home/lionel/src/olpc/sugarizer-server/sugarizer.js:31:26
    at /home/lionel/src/olpc/sugarizer-server/config/wait4db.js:20:6
    at result (/home/lionel/src/olpc/sugarizer-server/node_modules/mongodb/lib/utils.js:410:17)
    at executeCallback (/home/lionel/src/olpc/sugarizer-server/node_modules/mongodb/lib/utils.js:402:9)
    at err (/home/lionel/src/olpc/sugarizer-server/node_modules/mongodb/lib/operations/mongo_client_ops.js:286:5)
    at connectCallback (/home/lionel/src/olpc/sugarizer-server/node_modules/mongodb/lib/operations/mongo_client_ops.js:265:5)
    at server.connect (/home/lionel/src/olpc/sugarizer-server/node_modules/mongodb/lib/operations/mongo_client_ops.js:379:5)
    at Server.connectHandler (/home/lionel/src/olpc/sugarizer-server/node_modules/mongodb/lib/topologies/server.js:298:9)
    at Object.onceWrapper (events.js:286:20)
lionel@lionel-sugarizer:~/src/olpc/sugarizer-server$ 
```

llaske:
The current work around is to change `package.json` [line 16](https://github.com/llaske/sugarizer-server/blob/master/package.json#L16) from

            "express-validator": ">=5.3.1",

To

             "express-validator": "5.3.1",

PR #269, opened 2020-06-03
version for express-validator

makayabou:
As stated in https://github.com/llaske/sugarizer-server/issues/187, iit's needed to change the version of express-validator in packages.json, to avoid error on sugarizer start (id use it with docker-compose install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 		"express": "4.16.4",
 		"express-flash": "~0.0.2",
 		"express-session": "1.15.6",
-		"express-validator": "~5.3.1",
+		"express-validator": "5.3.1",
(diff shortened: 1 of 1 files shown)


llaske:
Thanks for your contribution but it's already fixed in https://github.com/llaske/sugarizer-server/commit/d23f85cc054b267e57f104e4d2d2d923076ecd42

In `package.json`:
- `>=5.3.1` means: any version greater than `5.3.1`, including version potentially incompatible with version `5.3.1`. It's why the issue was there.
- `~5.3.1` means: a version greater than `5.3.1` but compatible with version `5.3.1` so using `~` solve the issue.

PR #270, opened 2020-08-15
Configuration changes for Sugarizer School Portal

NikhilM98:
This commit contains the changes required to integrate Sugarizer Server with Sugarizer School Portal.
- Changes in the connection string to allow connection with mongodb replicaset.
- Dockerfile to generate Sugarizer Server image which will be used by [Sugarizer Chart](https://github.com/NikhilM98/sugarizer-chart) to create a container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Following is the typical content of Sugarizer Server settings file:
 	port = 27018
 	name = sugarizer
 	waitdb = 1
+        replicaset = false
 
 	[collections]
 	users = users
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:latest
```
llaske, line 1: Not sure what this Dockerfile is for?
Is it different from files in the `docker` directory?

NikhilM98: This Dockerfile is for generating a Docker image (hosted on DockerHub) which will be used to create Sugarizer Server containers on the Kubernetes Cluster. It's different from the Dockerfile in the docker directory.
- It does not require MongoDB like [here](https://github.com/llaske/sugarizer-server/blob/master/docker-compose.yml).
- It separately clones Sugarizer and Sugarizer Server and installs Sugarizer Server dependencies.

llaske: Okay, make sense. But to stay consistent, I suggest to move it in the `docker` directory and call it something like `Dockerfile-standalone` to make a difference with files `Dockerfile-mongodb` and `Dockerfile-server` already is this directory.

NikhilM98: @llaske I've moved it inside the `docker` directory.

llaske:
Nice. Thanks.

PR #277, opened 2021-03-30
Fixed Permission error in install.md

### LINKED ISSUE #276: Permission denied in installing Docker Compose
When using the script to install Docker compose given in [docs/install.md](https://github.com/llaske/sugarizer-server/blob/master/docs/install.md), "Permission denied" error is coming.

Screenshot from 2021-03-30 19-46-31

Line of Code on which error is coming:
```
curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" > /usr/local/bin/docker-compose
```

saurabhhere:
Fixed #276 

Updated docker compose install command as given [here](https://docs.docker.com/compose/install/):
```
curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -14,7 +14,7 @@ To run Sugarizer Server using Docker and Docker Compose:
 
 Install Docker Compose
 
-	curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" > /usr/local/bin/docker-compose
+	curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
 	chmod +x /usr/local/bin/docker-compose
 
 To install Docker Compose on ARM architectures (e.g. for the Raspberry Pi 3), the link above will not work.  You need to use [arm-compose](https://github.com/hypriot/arm-compose) instead.


saurabhhere:
@llaske Please review

llaske:
Not sure to understand why changing `>` into `-o` could fix the permission issue. If the user doesn't have permission to write in `/usr/local` directory, both syntax should fail.

Only adding `sudo` will change something.
BTW I'm not in favor to add `sudo` in documentation because the need to use it really depend of the operating system and the user permissions.
Plus, only responsible users should know when to use `sudo`.

NikhilM98:
I think @saurabhhere is using a machine which in which his user does not have a write permission on the `/usr/local/bin/` directory.

When he tries:
```
curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" > /usr/local/bin/docker-compose
```
It fails because he does not has the write permission.

When he tries:
```
sudo curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" > /usr/local/bin/docker-compose
```

It fails again. But why?

It's because the above command only runs the `curl` command with `sudo`. It uses `curl` and then redirects the output to `/usr/local/bin/docker-compose` using output redirection (`>`). The redirection is performed by the user which does not have the permission to write to that directory, it's not performed with sudo.

If he uses:
```
sudo curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
It works. It's because `curl` command is run with `sudo` just like the previous case however, this time curl performs the write operation on that directory.

There are several ways to perform output redirection, like (1) logging in as root user using or (2) running the shell with `sudo`.

llaske:
Thanks a lot for this explanation @NikhilM98 :+1: 
Just to be sure, are you in favor of the fix or not? :sweat_smile:

NikhilM98:
@llaske it's a small change with no negatives. It makes things a bit convenient for the users who need to use `sudo`. So, I'm in favour of this change.

llaske:
Thanks @saurabhhere , thanks @NikhilM98

Issue #289, opened 2022-02-18
Fix issues related to documentation in readmes (readme.md, docs/install.md)


ishubham21:
1. Under the section https://github.com/llaske/sugarizer-server/blob/master/docs/install.md#run-sugarizer-using-docker we need to write `sudo` before cURL and chmod commands because they are attempting to modify files in the root directory and need permissions. 
 2. Under the section https://github.com/llaske/sugarizer-server/blob/master/docs/install.md#run-sugarizer-on-linux, under "Install Sugarizer-server and sugarizer" sub-section, we need to update the command `cd /home/root/sugarizer-server` to `cd sugarizer-server` since the user is already in `/home/root` directory
 3. For some systems, `/home/root` directory may not be present and need to be made, so we need to add the command `cd /home && sudo mkdir root`
 4.  In the [main readme.md](https://github.com/llaske/sugarizer-server/blob/master/README.md) file, the instructions under https://github.com/llaske/sugarizer-server#running-sugarizer-server are not clear. It should also include a sentence `To access dashboard, follow the link - http://127.0.0.1:8080/`.

NikhilM98:
Yeah, you're right. Some instructions are ambiguous and the instructions for Mac and Linux are not consistent. 

> Under the section https://github.com/llaske/sugarizer-server/blob/master/docs/install.md#run-sugarizer-using-docker we need to write sudo before cURL and chmod commands because they are attempting to modify files in the root directory and need permissions.

In some parts of the documentation, it's assumed that the user is running as root user while in other parts, the user is non-root user. It'll be nice to make it consistent.


> Under the section https://github.com/llaske/sugarizer-server/blob/master/docs/install.md#run-sugarizer-on-linux, under "Install Sugarizer-server and sugarizer" sub-section, we need to update the command cd /home/root/sugarizer-server to cd sugarizer-server since the user is already in /home/root directory

It does not matters if the path is relative or fixed, both are correct. However, it'll be nice to make it consistent over the docs.

> For some systems, `/home/root` directory may not be present and need to be made, so we need to add the command `cd /home && sudo mkdir root`

Yeah, that's right. It's not good to assume that this path exists. Instead of forcing the user to install the Sugarizer stack at a fixed path, it'll be nice to make the instructions flexible like this:
```
git clone https://github.com/llaske/sugarizer
git clone https://github.com/llaske/sugarizer-server
cd sugarizer-server
npm install
```
```
[...]

Issue #287, opened 2022-02-03
Sugarizer-server port conflict


markknit:
I'm trying to install sugarizer-server on a Ubuntu 18.04 system using Nginx. I used the docker image, but when I try to finalize the install it fails due to a conflict on port 8080. There is already a service running that is bound to that port that I can't move. I tried changing the sugarizer.ini file to change the port, but it does not seem to have any effect on the install process - it still fails with the same error. What is the correct process for changing the sugarizer-server/docker port assignment, and then starting the sugarizer node? 

BTW: I also tried to install sugarizer-server without docker, but ran into a different problem - the npm superagent is deprecated. Is there a way to upgrade the superagent?

Thanks, Mark K

NikhilM98:
If you use Docker Compose to bring the stack up, then it should automatically map port 8080 to port 80 [here](https://github.com/llaske/sugarizer-server/blob/master/docker-compose.yml#L27).

Docker Compose instructions are given [here](https://github.com/llaske/sugarizer-server#running-sugarizer-server). You can change the port from port 80 to any other port in docker-compose.yaml.

NikhilM98:
If you want to update superagent then maybe you can use open package.json file, set `"superagent": "^7.0.0",` and try `npm install` again.

I have not tested the compatibility of the newer version, but given that superagent rarely has any [breaking changes](https://www.npmjs.com/package/superagent#upgrading-from-previous-versions), so you can give it a try.

markknit:
Thanks for the response. Changing the port in docker-compose.yml worked. However, there was another issue - the sugarizer-server folder is setup as root owned, so it wouldn't load. I changed it to www-data and it works fine now.

llaske:
> the sugarizer-server folder is setup as root owned
We're going to adapt the documentation to add recommendation regarding use of sudo.
Let's continue [here](https://github.com/llaske/sugarizer-server/issues/289).

PR #291, opened 2022-02-20
Issue #289: Fix issues related to documentation in readme files

ishubham21:
Fixes: https://github.com/llaske/sugarizer-server/issues/289

Changes - 
1. Add `sugarizer.log` in `.gitignore`
2. Add default address to access the server in `readme.md` file
3. Add a `Table of Contents` section in https://github.com/llaske/sugarizer-server/blob/master/docs/install.md
4. Update install instructions for Linux in https://github.com/llaske/sugarizer-server/blob/master/docs/install.md
5. Fix indentation under install instructions for macOS


ishubham21:
cc: @NikhilM98. Do let me know if any update is required. I've tested all the commands myself, they are working fine.

.gitignore
```diff
@@ -11,4 +11,4 @@ docker/db
 build/
 
 *.csv
-tmp/
+tmp/
```
NikhilM98, line 14: Is there any tmp directory getting created while Sugarizer-Server setup?
If no, then .gitignore does not needs to be modified.
If yes, then add a new line after this line.

README.md
```diff
@@ -17,6 +17,8 @@ The easiest way to run Sugarizer is to use Docker. To do that, type command line
 	sh generate-docker-compose.sh
 	docker-compose up -d
 
+By default, your sugarizer server will be accessible on http://127.0.0.1:8080 and your public IP.
```
NikhilM98, line 20: "and your public IP" does not makes sense. It would be more appropriate to specify that by default it will be available on port 8080.

Also, make use of markdown for links such as:
"By default, Sugarizer Server UI will be accessible on [http://127.0.0.1:8080](http://127.0.0.1:8080)"

docs/install.md
```diff
@@ -14,15 +23,20 @@ To run Sugarizer Server using Docker and Docker Compose:
 
 Install Docker Compose
 
-	curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" > /usr/local/bin/docker-compose
+	curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
NikhilM98, line 26: Can you replace 1.8.1 with 1.29.2?

docs/install.md
```diff
@@ -14,15 +23,20 @@ To run Sugarizer Server using Docker and Docker Compose:
...
 	chmod +x /usr/local/bin/docker-compose
 
+If you have problems installing with `curl`, you can use `pip` instead: `pip install -U docker-compose`
```
NikhilM98, line 29: There are certain conditions while installing docker-compose with pip which needs to be fulfilled.
See: https://docs.docker.com/compose/install/#alternative-install-options

It's misleading to suggest pip as the installation method without highlighting the recommended requirements.

For example, the users are not aware that "If you install using pip, we recommend that you use a [virtualenv](https://virtualenv.pypa.io/en/latest/) because many operating systems have python system packages that conflict with docker-compose dependencies."

So, instead of suggesting the use of pip to install docker compose, it's better to add a reference to this (https://docs.docker.com/compose/install/) documentation so that user can install correct version of docker-compose based on their OS.

docs/install.md
```diff
@@ -36,28 +50,34 @@ Your Sugarizer server will start automatically and will be accessible on http://
...
+    sudo apt update
+	sudo apt install nodejs
 
-**Install MongoDB**: Install MongoDB (2.6+). Don't forget to create a /data/db directory to store databases. See [here](http://www.mongodb.org/ "here") more information.
+**Install MongoDB**: Install MongoDB (2.6+). Don't forget to create a /data/db directory to store databases. See [here](https://docs.mongodb.com/manual/installation/ "here") for more information.
 
     sudo apt-get install mongodb
     sudo mkdir -p /data/db
 
+**Install Git**: Install Git (2.17+). See [here](https://github.com/git-guides/install-git) for more information.
```
NikhilM98, line 63: It's fine to assume that the user have Git installed. Also, Git is not something which is required to run Sugarizer Server. So this part is not needed.

docs/install.md
```diff
@@ -36,28 +50,34 @@ Your Sugarizer server will start automatically and will be accessible on http://
 To run Sugarizer Server **without Docker**, follow the step behind. Commands are shown from a new Debian Linux machine and could be different for other Linux distribution or for an already installed machine:
 
 
-**Install Node.js**: Install Node.js (6+) and npm to manage packages. See [here](http://nodejs.org/ "here") more information.
+**Install Node.js**: Install Node.js (6+) and npm to manage packages. See [here](http://nodejs.org/ "here") for more information.
 
-    sudo apt-get install nodejs
+    sudo apt update
```
NikhilM98, line 55: Use of api and apt-get is inconsistent in the documentation.

docs/install.md
```diff
@@ -36,28 +50,34 @@ Your Sugarizer server will start automatically and will be accessible on http://
...
+	sudo apt install nodejs
```
NikhilM98, line 56: Fix indentation here.

docs/install.md
```diff
@@ -36,28 +50,34 @@ Your Sugarizer server will start automatically and will be accessible on http://
...
+    npm install
 
 **Run MongoDB and Sugarizer Server**:Run mongo daemon and Sugarizer a background process.
 
-    sudo mongod --fork --port 27018 --logpath /home/root/mongo.log
-    sudo nohup node sugarizer.js > /home/root/sugarizer.log &
+    sudo mongod --fork --port 27018 --logpath ~/mongo.log
+    nohup node sugarizer.js > ~/sugarizer.log
+
+By default, your sugarizer server will be accessible on http://127.0.0.1:8080 and your public IP.
```
NikhilM98, line 80: https://github.com/llaske/sugarizer-server/pull/291/files#r825305718

docs/install.md
```diff
@@ -68,54 +88,55 @@ To run Sugarizer Server **without Docker**, follow the step behind. Commands are
...
 **Install Sugarizer Client and Server**
 
-        brew install git
-        cd desktop
-        git clone https://github.com/llaske/sugarizer
-        git clone https://github.com/llaske/sugarizer-server
-        cd sugarizer-server
-        npm install
+    brew install git
+    cd desktop
```
NikhilM98, line 128: It does not seems necessary to `cd desktop`

docs/install.md
```diff
@@ -68,54 +88,55 @@ To run Sugarizer Server **without Docker**, follow the step behind. Commands are
...
+    npm install
 
 **Run MongoDB and Sugarizer Server**:Run mongod and Sugarizer.
 
-        mongod
-        node sugarizer.js
+    mongod
+    node sugarizer.js
 
+By default, your sugarizer server will be accessible on http://127.0.0.1:8080 and your public IP.
```
NikhilM98, line 139: https://github.com/llaske/sugarizer-server/pull/291/files#r825305718

docs/install.md
```diff
@@ -36,28 +50,34 @@ Your Sugarizer server will start automatically and will be accessible on http://
...
 **Install Sugarizer Client and Server**: If need, you could update sugarizer-server/sugarizer.ini file (update port for web, mongodb or presence, see below)
 
-    sudo apt-get install git
-    cd /home/root
-    sudo git clone https://github.com/llaske/sugarizer
-    sudo git clone https://github.com/llaske/sugarizer-server
-    cd /home/root/sugarizer-server
-    sudo npm install
+    git clone https://github.com/llaske/sugarizer
+    git clone https://github.com/llaske/sugarizer-server
```
NikhilM98, line 71: This change will fix [this](https://github.com/llaske/sugarizer-server/issues/287#issuecomment-1048228896) issue.

PR #311, opened 2022-03-19
Fix Lint errors

dhwaj1902:
While running ``npm run lint``, there are some errors:
before

Afer fixing:
after


dashboard/controller/users/enable2FA.js
```diff
@@ -14,7 +14,7 @@ module.exports = function enable2FA(req, res) {
 
 	if (req.method == 'POST') {
 
-		req.assert('tokenentry', common.l10n.get('TokenAtLeast', {min_token_size})).len(min_token_size);
+		req.assert('tokenentry', common.l10n.get('TokenAtLeast', min_token_size)).len(min_token_size);
```
NikhilM98, line 17: Have you tested your changes? I think this change will break l10n translation argument

dhwaj1902: Yes, but it is giving the error in lint testing, can I add this file in lint ignore?

NikhilM98: We cannot add code which is managed by us to lint ignore. There would be no point of adding linter to the project if we do that. However, it should be fine to update the rules of eslint so that it does not detects intentional code as bug.


But in this case, just using this would fix the linter issue:
```
{ min_token_size: min_token_size }
```

Btw, have you verified if your change breaks l10n or not?

dhwaj1902: Yes @NikhilM98, due to this change it is breaking the 2-factor authorization, thanks! for the help!

llaske:
I can't merge a PR in master branch. @dhwaj1902 could you send it to the dev branch?

dhwaj1902:
> I can't merge a PR in master branch. @dhwaj1902 could you send it to the dev branch?

Done.

llaske:
Nice. Thanks.

PR #312, opened 2022-03-20
fix #292

### LINKED ISSUE #292: Popups in Dashboard seem to be broken
Screenshot from 2022-02-20 22-11-38
The popups in the Dashboard looks like they are not able to retrieve content from the locales.ini file. 

One thing to note here is that I was able to fix it via refreshing it 3-4 times but it again broke while I refreshed it for the 5th time.

Steps to reproduce -
1. Log in to the Dashboard of Sugarizer Server. 
2. Clear all cache and site data, make sure that you've emptied the localStorage so that popovers can be loaded. 
3. Log in once again.
4. Head on `Statistics View` and try refreshing 3-4 times while the popup is open. 

I'd love to work on this but I wanted to make sure if that everyone is getting this problem.

mohantechnology:


mohantechnology:
fix #292

llaske:
No PR can be merged from master branch.

mohantechnology:
to fix the issue i have also made some changes  "dashboard\public\js\l10n.js", as it is a library and  changing it may cause some other problems . so another way to  solve this issue is to add attribute name to all element which uses   'document.webL10n.get( )' 

so in  below screenshots i have added attribute only 'prev' button 

https://user-images.githubusercontent.com/71864565/161443943-9a2143b8-c615-4a29-820c-ebf8d551b0df.mp4

if this is correct then i would like to make PR after adding attribute to other elememts also

Issue #313, opened 2022-03-23
filtering out data while exporting user accounts from admin dashboard


saga2405:
the button `export users` in the admin dashboard(`http://127.0.0.1:8080/dashboard/users`) exports all types of user accounts in the form of CSV file regardless of the type(`student, teacher, admin`) we've selected
the downloaded file includes all three types of accounts - admin, teacher, student
We can filter out the data and export only those type of account which is selected in the query 
for example: If I'm viewing student accounts and click on export users - the downloaded csv file should only contain users of type student, the same goes for teacher and admin type
should i fix this?

saga2405:
@llaske can you please give your inputs

llaske:
You're right it make sense to use filter (both user type and user name) in filtering.

saga2405:
@llaske @NikhilM98 I've made a request for this. can you please check

parteekcoder:
hey @NikhilM98 @llaske Can I work upon this

NikhilM98:
@parteekcoder There is someone already working on this issue. Let's wait for them to respond.

PR #318, opened 2022-03-27
filtering users by name and account type while exporting

saga0p:
Now the users are getting filtered according to account type and name when we export users, rather than exporting all the accounts , only the filtered accounts are getting downloaded


saga2405:
issue #313

saga2405:
@llaske @NikhilM98 can you please review this request (#313 )

NikhilM98:
There are several changes which are not relevant to the issue (comments, unwanted spaces). Can you please remove them?

Also, if you look at the git diff, you'll find some "end of line" errors, can you please get rid of them too?

saga2405:
sure, I'll do that

saga2405:
@NikhilM98 done

add-admin.sh
```diff
@@ -2,4 +2,4 @@ if ! [ $# -eq "3" ] ; then echo "You should provide user, password and URL" ; ex
 newuser=$1
 newpassword=$2
 newobject="user=%7B%22name%22%3A%22${newuser}%22%2C%22role%22%3A%22admin%22%2C%22password%22%3A%22${newpassword}%22%2C%22language%22%3A%22en%22%7D"
-curl -k -i -H "Content-Type: application/x-www-form-urlencoded" -X POST -d "${newobject}" $3
+curl -k -i -H "Content-Type: application/x-www-form-urlencoded" -X POST -d "${newobject}" $3
```
llaske, line 5: Remove this change from the PR

docs/install.md
```diff
@@ -14,7 +14,7 @@ To run Sugarizer Server using Docker and Docker Compose:
 
 Install Docker Compose
 
-	curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" > /usr/local/bin/docker-compose
+	curl -L "https://github.com/docker/compose/releases/download/1.8.1/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
```
llaske, line 17: Remove this change from the PR

env/sugarizer.ini
```diff
@@ -1,3 +1,4 @@
+version = 1.4.0
```
llaske, line 1: Remove this change from the PR

package-lock.json
```diff
@@ -1,12806 +1,8 @@
 {
 	"name": "sugarizer-server",
 	"version": "1.4.0",
-	"lockfileVersion": 2,
+	"lockfileVersion": 1,
```
llaske, line 4: Remove changes in this file from the PR

llaske (changes requested):
Good job. It works and it's useful.
BTW it's no longer possible to export every users which is require.
I suggest to add an option to do that in the combo to allow Students/Teacher/Admin + All.

See also my comments below and remove comments on every line of your change.

llaske:
Closed due to lack of activity.

Issue #348, opened 2023-02-08
Fix required in sidebar to achieve consistency


mayank190801:
As you can see in the video - 
 - Between window-width from 768px to 921px the logout and profile button doesn't appear, which creates inconsistency in the overall sidebar. 

A need for creating consistency till the max-width of 922px could be achieved in the sidebar.
In which there could be no dropdown option, but only logout and profile button appears as soon as the navbar collapses.

https://user-images.githubusercontent.com/54149585/217568863-9d886a68-24cc-4c1e-9a22-74ce62d93541.mov

mayank190801:
@llaske I would like to know if you think this issue needs some fix?

llaske:
You're right, it could be improved.

llaske:
Fixed in https://github.com/llaske/sugarizer-server/pull/353

PR #349, opened 2023-02-09
Fixed the required changes in sidebar for better user experience

mayank190801:
Fix for the issue #348 

I have made 
- Made the user dropdown for profile feature and logout feature "hidden" once the width is less than 992px.
- Made the logout button visible once it is under the width of 992px. (earlier it was only for less than 768px)
- Removed box shadow from language selector for much more cleaner look (because it was only there when the width was less than 768px and not 992px).
- Also reduced some margin between logout button and language selector for better experience.

Below I have attached a video of how it looks now. 

https://user-images.githubusercontent.com/54149585/217747260-02657341-806c-4b61-815b-c794b9e13eee.mov

### DIFF
--- a/dashboard/public/css/main.css
+++ b/dashboard/public/css/main.css
@@ -846,6 +846,12 @@
   display: none !important;
 }
 
+@media (max-width: 991px) {
+  .hide-desk-el {
+    display: block !important;
+  }
+}
+
 @media (max-width: 768px) {
   .pull-right-mobinterrupt {
     float: none !important;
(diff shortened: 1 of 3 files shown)


llaske:
Sounds good but the `package-lock.json`, not related with this fix, should not be changed.

mayank190801:
Thank you for pointing that out. I missed that. It would have happened automatically while I was installing the packages. I will fix that.

mayank190801:
I have created a new pull request as I did not want to pollute the history. I will be closing this one.

Issue #351, opened 2023-02-10
Activities list not completely centered


mayank190801:
-- Detailed Issue -- 

- There have been some few pixels space left inside each div, inside of list element of Activites.
- Due to which the text, icon, number is not completely centered with respect to the list .
- There is a more space in the bottom and less from the top.

![Space_Beneath](https://user-images.githubusercontent.com/54149585/218071054-2aef1f32-92ed-4df9-863d-079c0806c49a.png)

llaske:
Really not a visible issue. I don't think any user have detected that.

mayank190801:
Should I close this issue then?

llaske:
> Should I close this issue then?

Yes

PR #352, opened 2023-02-10
Centered the divs inside list of Activities

mayank190801:
Fix for the issue #351 

-- Detailed Fix -- 

- I have first set the height of divs equal to height of the list. (to finish that extra space below each div)
- Then using flexbox properties I have centered everything inside the div, height and width wise. 

Screenshot 2023-02-10 at 4 04 44 PM


mayank190801:
@llaske Kindly review it and let me know if any changes are required.

PR #353, opened 2023-02-12
fixed sidebar logout option

mayank190801:
Fix for the issue #348 

Changes I have made :- 

- Made the user dropdown for profile feature and logout feature "hidden" once the width is less than 992px.
- Made the logout button visible once it is under the width of 992px. (earlier it was only for less than 768px)
- Removed box shadow from language selector for much more cleaner look (because it was only there when the width was less than 768px and not 992px).
- Also reduced some margin between logout button and language selector for better experience.
Below I have attached a video of how it looks now.

https://user-images.githubusercontent.com/54149585/218292935-86fb2385-6ad9-4bfe-a4af-ee0d5ef856e7.mov

### DIFF
--- a/dashboard/public/css/main.css
+++ b/dashboard/public/css/main.css
@@ -846,6 +846,12 @@
   display: none !important;
 }
 
+@media (max-width: 991px) {
+  .hide-desk-el {
+    display: block !important;
+  }
+}
+
 @media (max-width: 768px) {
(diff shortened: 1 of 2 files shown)


mayank190801:
@llaske Please review it and let me know if any changes are required.

llaske:
Nice. Thanks for this fix.

Issue #360, opened 2023-02-16
Update the Popover box 


mayank190801:
Need For Enhancement 

- I have found that Sugarizer and Sugarizer-Server used to have similar types of Popover boxes before. Recently inside the dev-branch of Sugarizer, the popover box is updated and has a much better look and feel to it.

- I would like to work on this enhancement by updating the Sugarizer-Server popover box as well.

-- Updated Popover Box inside Sugarizer (dev branch) -- 
<img width="281" alt="Screenshot 2023-02-16 at 10 39 21 AM" src="https://user-images.githubusercontent.com/54149585/219280080-9161c2ba-c6a3-4010-b13b-4e219a19149b.png">

-- Current Popover Box inside Sugarizer-Server (dev branch) -- 
<img width="327" alt="Screenshot 2023-02-16 at 10 32 13 AM" src="https://user-images.githubusercontent.com/54149585/219280058-33c7766f-c10c-4d18-9a2b-d2617f1dceaf.png">

mayank190801:
@llaske What are views regarding it? 
I have also made a pull request regarding this enhancement.

mayank190801:
@llaske Is there a community group (slack, discord etc.) for people contributing to Sugarizer or Sugarizer-Server?

llaske:
It's not related to the popover box, it comes from the library used for tutorial.
See [here](https://github.com/llaske/sugarizer/issues/1090) for more.
It will be a task for the GSoC to migrate.

mayank190801:
Okay. So can I try to resolve it or should I leave it for GSoC?

llaske:
Yes you can start working on it. I've opened a dedicated issue here https://github.com/llaske/sugarizer-server/issues/363.
You can close the current issue.

mayank190801:
Okay. Thank you.

PR #361, opened 2023-02-16
Updated the popover box

mayank190801:
Fixes issue #360 

-- Detailed changes -- 

- Remove the end button from the bottom, and used only the cross-icon portion of it in the title.
- Written some extra divs and classes for adjusting title and icon in the same line.
- Written some extra css for styling the title div.
- Removed the dash between the prev and next by changing the color to white.
- Centered prev and next text by reducing some upper space.
- Reduced some space between button and contents.
- Increased the max-width from 276px to 300px of the box.

Before 
Screenshot 2023-02-16 at 10 39 03 AM

After 
Screenshot 2023-02-16 at 12 15 31 PM

### DIFF
--- a/dashboard/public/css/main.css
+++ b/dashboard/public/css/main.css
@@ -733,7 +733,7 @@
 }
 
 .icon-tutorial-text {
-	margin-top: 7px;
+	margin-top: 5px;
 	padding-left: 38px;
 	padding-right: 38px;
   display: inline-block;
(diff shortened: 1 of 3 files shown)


mayank190801:
Fixes issue #360

llaske:
It's not the right way to solve the problem. See my comment on your issue..

Issue #371, opened 2023-02-22
Error while requesting


parteekcoder:
Reproduce

1. Start the server
2. Make a request

The function where JSON.parse() is used having req object in the parse function, It will give following error

![image](https://user-images.githubusercontent.com/100110395/220733917-ac99d16b-978c-4aef-bc4b-8127c43d9509.png)

As this is useless , we don't need to use it here 
Because we are using `bodyParser` to convert req body

codebloded:
@parteekcoder 
Kindly check your JSON data you passed in body from Postman. It accepts whole data as a string !
console it and check what you have passed in body !

parteekcoder:
> @parteekcoder Kindly check your JSON data you passed in body from Postman. It accepts whole data as a string ! console it and check what you have passed in body !

@codebloded yeah I have checked that But still is it failing

codebloded:
Give me a screenshot of body that you have passed in the request from postmam

parteekcoder:
![image](https://user-images.githubusercontent.com/100110395/220903676-51da2e88-d9f0-46d1-9854-5630a25c8194.png)

I passed json as well as stringified data

codebloded:
> ![image](https://user-images.githubusercontent.com/100110395/220903676-51da2e88-d9f0-46d1-9854-5630a25c8194.png)
> 
> I passed json as well as stringified data

This is good but I'm expecting the screenshot of body that you have passed from the postman!
Are you using postman?

parteekcoder:
> > ![image](https://user-images.githubusercontent.com/100110395/220903676-51da2e88-d9f0-46d1-9854-5630a25c8194.png)
> > I passed json as well as stringified data
> 
> This is good but I'm expecting the screenshot of body that you have passed from the postman! Are you using postman?

![image](https://user-images.githubusercontent.com/100110395/220906671-8694f9e3-9166-4418-811a-92fcbefcb800.png)

parteekcoder:
> Reproduce
> 
> 1. Start the server
> 2. Make a request
> 
> The function where JSON.parse() is used having req object in the parse function, It will give following error
> 
> ![image](https://user-images.githubusercontent.com/100110395/220733917-ac99d16b-978c-4aef-bc4b-8127c43d9509.png)
> 
> As this is useless , we don't need to use it here Because we are using `bodyParser` to convert req body

hey @llaske please see this error it is due to the JSON.parse() we are using

parteekcoder:
[...]

PR #372, opened 2023-02-22
Some error

parteekcoder:
issue #371


llaske:
Please @parteekcoder, do not submit a PR until we've reproduce your issue. We need to ensure that's really an issue.

parteekcoder:
hey @llaske please verify issue as I got it many times

Your suggestions will be admired

llaske:
Not an error

PR #374, opened 2023-02-23
Refer to the MongoDB container in the migrate to v1.5.0 docs

cperreault:
I think that it was wanted to refer to the MongoDB container instead of the `server container. That's what what I did to successfully migrate from v1.3.0 to v1.5.0.


llaske:
Oops, you're right.
Thank for your contribution.

Issue #364, opened 2023-02-18
Next page button in users/assignment view don't take into account search criteria


llaske:
In the Users view and in the Assignment view, the next button don't take into account search criteria.

See the problem below (for the Users view):

![cNyOEOR9W8](https://user-images.githubusercontent.com/981841/219852386-6479152d-6508-43cd-a5f5-fdff9b6d4986.gif)

arian0zen:
hey @llaske , was asking that I want to work on this issue.. but I am not familiar on how to set up the developer enviornment for this sugarizer server, can you help me a bit with this ? or am I missing something obvious here ?

update:  10:38 pm 25th feb
oh I am sorry, I was confused where is the repo for this sugarizer and I was looking in the main sugarlabs repo... so do I clone this repo llaske/sugarizer-server and set up the enviornment ?

another update query: 11:17 pm 25th feb
I was carefully looking at the cource code on different folders from both sugarizer and sugarizer server.. I felt i am familiar with the .js codes and also I am confident I can work on a few of features raised in the issues..
but, Unfortunately I am not at all familiar with the docker and staff so I am finding it a bit difficult set up to start to code.. I really really need help on thisss. hey @NikhilM98  I am mentioning you here as well, if you can help me with an way around, thankkk you, looking forward :)

llaske:
@arian0zen just follow running instructions [here](https://github.com/llaske/sugarizer-server/tree/dev#running-sugarizer-server) and you will have a running environment

arian0zen:
do I have keep to both sugarizer and sugarizer-server folder in the same directory ? what should be the heierarchy ?

llaske:
Not in the same directory, side by side. Clone commands create a `sugarizer` and a `sugarizer-server` directory at the same level.

arian0zen:
hey, I kinda got the set up ready, though `sh generate-docker-compose.sh` this command is not working for me, I am getting this error, (i am on windows 10)
![image](https://user-images.githubusercontent.com/68517592/221391635-d5a211d8-4e39-4bb8-9534-96b79d32737b.png)

why is this ? 
though when I am using `nodemon sugarizer.js` I see the local host working on 8080.

So, I was asking should I stop worrying about why docker compose is not working in my setup and start working on the features ? or `sh generate-docker-compose.sh` this command is necessary ?

- do I have to install `docker` first like this ? is it necessary for development ? 
[...]

PR #379, opened 2023-03-04
Users next page button fix

Asymmentric:
##### Fix for the issue https://github.com/llaske/sugarizer-server/issues/364

Complete: 
- Fixed 'Next page >>' and '<< Previous page' buttons on `dashboard/users`  which now considers the search criteria while displaying user details
- form on `/dashboard/users` has 'username' as query parameter, so replaced 'q' query parameters by 'username' throughout 

In Progress:
- 'Next page >>' for assignments


NikhilM98:
Can you clean this PR?
- Get rid of commits which are not yours.
- Remove unnecessary comments and white space from the code.

Asymmentric:
hi @NikhilM98 
i tried cleaning it. Please let me know if it can be improved

docs/migrate.md
```diff
@@ -46,7 +46,7 @@ First identify the id for the MongoDB container by running.
 docker ps
 ```
 
-The id is the one on the line named `sugarizer-server_server`.
+The id is the one on the line named `sugarizer-server_mongodb`.
```
NikhilM98, line 49: Create a different PR to address this typo.

api/controller/users.js
```diff
@@ -559,11 +554,10 @@ function addQuery(filter, params, query, default_val) {
 
 	//check default case
 	query = query || {};
-
 	//validate
 	if (typeof params[filter] != "undefined" && typeof params[filter] === "string") {
-
-		if (filter == 'q') {
+		
+		if(filter=='username'){
```
NikhilM98: There is no point in replacing `q` with `username`. `q` is commonly used for search operations.

api/controller/users.js
```diff
@@ -559,11 +554,10 @@ function addQuery(filter, params, query, default_val) {
 
 	//check default case
 	query = query || {};
-
```
NikhilM98, line 562: These delete line changes are unnecessary.

api/controller/users.js
```diff
@@ -45,8 +45,7 @@ exports.init = function(settings, database) {
  * @apiParam {String} id Unique user id
  * 
  * @apiSuccess {String} _id Unique user id
- * @apiSuccess {String} name Unique user name
- * @apiSuccess {String} role User role (student or admin)
+ * @apiSuccessearchs {String} role User role (student or admin)
  * @apiSuccess {Object} color Buddy color
```
NikhilM98, line 49: What's the purpose of this change?

dashboard/controller/users/index.js
```diff
@@ -71,7 +70,7 @@ exports.index = function(req, res) {
 
 				// get classrooms list
 				getClassrooms(req, function(classrooms){
-	
+
```
NikhilM98, line 73: The changes with extra whitespaces are unnecessary.

NikhilM98 (changes requested):

NikhilM98:
In this PR I can see that you have changed some variable names, but how does it fixes the next page button issue?

Asymmentric:
Suppose when the SHOW RESULTS button is clicked it requests   `_dashboard/users?username=n&role=student_` here it was fetching the results good and it was being resolved as  `query['q']=req.query.username`  but the _next_page_ links which were being created by `formPaginatedUrl()` looked something like  `_dashboard/users?q=n&role=student&offset=10&limit=10_` 

The problem was due to `req.query.q` not being handled there so the search criteria was not being considered.

NikhilM98:
I just tested the issue. I see what you mean. However, it may be better if the keys like `username`, `name`, and `classroom` are reserved for exact string matching. For searching with partial string matching, it may be better to use `q`.

What do you think about doing something like [this](https://github.com/NikhilM98/sugarizer-server/commit/642260896f1e882c7fbcd685c6b36851dbfee0c2)?

Asymmentric:
> What do you think about doing something like [this](https://github.com/NikhilM98/sugarizer-server/commit/642260896f1e882c7fbcd685c6b36851dbfee0c2)?

True, that is a fair approach but rather than changing element names in the ejs file can this be a possible solution 

By handling _req.query.q_

`if (req.query.username != '' || req.query.q!='') {
 query['q'] = req.query.username || req.query.q;
 }`

`if (req.query.classroom != '' || req.query.q!='') {
 query['q'] = req.query.classroom || req.query.q;
 }`

NikhilM98:
Yes this is a possible solution too, but it is prone to unexpected behavior, for example if in future both variables are defined. Also, it may be confusing to use two variables for the same parameter.

Asymmentric:
Yes sure it might make it inconsistent

NikhilM98:
Can you make your changes consistent with the suggested ones and squash your commits

Asymmentric:
@NikhilM98 I've made the changes. Let know if any improvements

NikhilM98:
There are several changes in this PR which are not relevant to this issue: [changes](https://github.com/llaske/sugarizer-server/pull/379/files).

Can you remove the unnecessary changes and squash your commits into one or two commits?

kartikaysaxena:
Can I contribute to the same?

NikhilM98:
@kartikaysaxena the solution is already finalized and coded. This tasks doesn't require collaboration between multiple contributions for now. Feel free to experiment with Sugarizer Server and report any issues you find in the application.

llaske:
@Asymmentric could you close one of the two PR?

Issue #365, opened 2023-02-18
Improve create assignment form: add a search activity in journal feature 


llaske:
When a teacher want to create an assignment, he need to choose an activity instance in its Journal. 
A popup is displayed with 10 latest activities :

![image](https://user-images.githubusercontent.com/981841/219852502-14bdfa43-f9eb-4c8e-a551-7b00ca3ca704.png)

The idea is to replace this popup by a popup similar to the one used to look for students in the Journal view:

![image](https://user-images.githubusercontent.com/981841/219852657-5d1ea151-cbce-45a8-8bb4-595222a29dd6.png)

The new popup should:
- Allow to search for any activity in the teacher's Journal (not only the 10 latest)
- Show a preview of the activity icon

parteekcoder:
hey @llaske I am not getting this type of interface to add assignment, could you please help me

PR #386, opened 2023-03-07
Update addAssignment.js

parteekcoder:
issue #365 

Now all the activities are shown not just 10

### DIFF
--- a/dashboard/controller/assignments/addAssignment.js
+++ b/dashboard/controller/assignments/addAssignment.js
@@ -13,7 +13,7 @@ module.exports = function addAssignment(req, res) {
 	// reinit l10n and momemt with locale
 	common.reinitLocale(req);
 	var query = {
-		limit: (req.query.limit ? req.query.limit : 10),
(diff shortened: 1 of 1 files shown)


parteekcoder:
@NikhilM98 @llaske I want to ask that we want this `icon` to be present inside the selection input or outside the selection input(now implemented)

image

parteekcoder:
@NikhilM98 please give your suggestion , so that I get to know about the functionality required

parteekcoder:
hey @NikhilM98 @llaske  Now it has as many journals as user can there no restriction like 10 , please review and I would appreciate the changes you prefer

llaske:
> I want to ask that we want this icon to be present inside the selection input or outside the selection input(now implemented)

As mentioned in the issue, we want a popup similar to the one used to look for students in the Journal view. So with the icon inside the selection. Once selected the icon will be outside as today.

> Now it has as many journals as user can there no restriction like 10 

Hmmm. The idea was to make a smarter implementation, like a search field. Not to hardcode the limit.

